PR #42, opened 2023-07-04
Modify shard files to include dedup hint refs to other shards.

hoytak:
This PR adds shard dedup count tracking the shard file manager, which counts, for current dedup operations, how many times each on-disk shard gets referenced.  When shards are flushed to disk, this information is stored in a section in the shard as a list of hashes of other shards and dedup counts.  The format for the shard files regarding the intershard reference section is fully backwards and forwards compatible.

The tricky case here involves handling a large file that ends up generating many shards in one session.   Ideally, we would like to be able to dedup this file; however, the references are all to shards with in the session, which may be merged later on.   Without tracking intra-session references, subsequent small modifications to a large file will not see much dedup.  

This PR solves this, including this use case, by: 

1.  Dedup counts against written shards are tracked, even in the same session.  These must be previously written shards. 
2.  The format of the staged shards now includes a counter.  Now staged shards are S####-HASH.mdb.   Note that any references in to one shard will always be to a shard with a lower counter number or a permanent (in CAS / cache) shard. 
2.  On upload, the merge process only merges sequential shards.  During the sequential merges, a hash map of remapped shard hashes is maintained.  The shard hint section is remapped on merge, or for staged shards appearing after a merge, the remap occurs.  
3.  Shard registering is gated on all shards successfully uploading instead of just the shard itself successfully uploading.  Thus all references in any registered shard will always be in CAS.


hoytak:
Succeeded by https://github.com/xetdata/xet-core/pull/66

PR #46, opened 2023-07-06
Added tool to verify MerkleDB shard.

hoytak:
For some reason, some shards appear to not parse correctly on the shard server, hitting an EOF when asked to read the file reconstruction range.   This PR is an attempt to understand why.

Usage: 

git xet merkledb verify <path>


hoytak:
Put in with a different version in https://github.com/xetdata/xet-core/pull/61

PR #53, opened 2023-07-17
Treat many chunk lookup collisions as info instead of error

seanses:
Fix https://github.com/xetdata/xethub/issues/3487


hoytak:
Hmm.  This would fix the acute problem, but the issue is that true random collisions are not deduplicated against.  I think ideally we should limit this number to the max consecutive chunks here on write.  Or simply make that not an error if more than 8 are present on lookup -- i.e. only a subset are returned.

hoytak (approved):
Nit: Can likely drop the TruncatedHashCollisionError from the error enum as this is the only place it's generated.   But looks good.

PR #54, opened 2023-07-18
Write transaction temp session dir

seanses:
Since pyxet will create multiple write transactions and they run in parallel, a temporary session dir should be created for each write transaction, instead of each xet repo.


hoytak (approved):
One small issue that should be fixed with this PR.  The Drop trait on shard_file_manager tries to write to the session_dir.  However, because it's a temp directory, that may fail based on which one goes out of scope first, or it gets recreated on demand.  So we should add a check to the drop() method to only flush to the session directory if the directory exists.

hoytak (changes requested):
See comment above

hoytak (approved):

PR #55, opened 2023-07-18
Allows dedup against previous chunks within the same CAS block on clean.

hoytak:
Dedup against previous chunks in the same CAS block, preventing a sequence of identical chunks to fill up a CAS block and cause lookup table irregularities.

### DIFF
--- a/rust/gitxetcore/src/data_processing_v2.rs
+++ b/rust/gitxetcore/src/data_processing_v2.rs
@@ -1,4 +1,5 @@
 use std::clone::Clone;
+use std::collections::HashMap;
 use std::ffi::OsStr;
 
 use std::ops::DerefMut;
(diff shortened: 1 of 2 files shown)


seanses (commented):
LGTM. 
As I read the code there's one line I don't quite understand, though it isn't part of this PR. On [this line](https://github.com/xetdata/xet-core/blame/a77e1d3e15a35db625bcdf36e83dc0314a1ef791/rust/gitxetcore/src/data_processing_v2.rs#L330) `chunk.hash` is compared with `file_info.last().unwrap().cas_hash`. IIUC `file_info[i].cas_hash` should either be `MerkleHash:default()` if the corresponding CAS block is not registered yet, or some valid CAS block hash. So when can it equal to a chunk hash?

seanses (approved):

hoytak:
Good catch on the hash comparison.  Will fix.

PR #58, opened 2023-07-19
Remove libmagic dependency

znation:
Reasons for removing libmagic are numerous, but the main ones are:
* It's very slow
* It lies about file types
* It doesn't work on Windows

This PR replaces file type detection using libmagic with a file-extension based lookup (see `file_types.rs`). The lookup table is from file extension to mime type and friendly name, and uses [phf](https://docs.rs/phf/latest/phf/index.html) for efficient compile-time construction of the hash table.


rust/libmagic/src/file_types.rs
```diff
@@ -0,0 +1,716 @@
...
+    "xwd" => FileTypeInfo { friendly_type: "X Window Dump", mime_type: "image/x-xwindowdump"},
+    "der" => FileTypeInfo { friendly_type: "X.509 Certificate", mime_type: "application/x-x509-ca-cert"},
+    "fig" => FileTypeInfo { friendly_type: "Xfig", mime_type: "application/x-xfig"},
+    "xhtml" => FileTypeInfo { friendly_type: "XHTML - The Extensible HyperText Markup Language", mime_type: "application/xhtml+xml"},
+    "xml" => FileTypeInfo { friendly_type: "XML - Extensible Markup Language", mime_type: "application/xml"},
+    "xdf" => FileTypeInfo { friendly_type: "XML Configuration Access Protocol - XCAP Diff", mime_type: "application/xcap-diff+xml"},
+    "xenc" => FileTypeInfo { friendly_type: "XML Encryption Syntax and Processing", mime_type: "application/xenc+xml"},
+    "xer" => FileTypeInfo { friendly_type: "XML Patch Framework", mime_type: "application/patch-ops-error+xml"},
+    "rl" => FileTypeInfo { friendly_type: "XML Resource Lists", mime_type: "application/resource-lists+xml"},
+    "rs" => FileTypeInfo { friendly_type: "XML Resource Lists", mime_type: "application/rls-services+xml"},
```
hoytak, line 702: Not sure this one is right... Could this be an older list?

hoytak:
I like it, +100.  
- I worry that the list is a bit dated.  Maybe it'd be worth writing a quick script to scrape https://en.wikipedia.org/wiki/List_of_filename_extensions_(A–E)#A,  etc. to use that?  We can keep it up to date that way.
- Let's do another PR after this to just replace "libmagic" with filetype or something like that.

seanses:
> I like it, +100.
> 
> * I worry that the list is a bit dated.  Maybe it'd be worth writing a quick script to scrape [https://en.wikipedia.org/wiki/List_of_filename_extensions_(A–E)#A](https://en.wikipedia.org/wiki/List_of_filename_extensions_(A%E2%80%93E)#A),  etc. to use that?  We can keep it up to date that way.
> * Let's do another PR after this to just replace "libmagic" with filetype or something like that.

There are extension collisions in that wiki list. Definitely need some manual screening there.

jgodlew (approved):
LGTM! :shipit:

rust/libmagic/src/file_types.rs
znation: Very possible. Also possible it's simply a file extension clash. Lots of file extensions get re-used and I think Rust probably picked one that was already in "common" use prior to its existence. :(

znation: I'll fix that one since I think it's infinitely more likely people will have Rust code with .rs than XML resource lists with .rs.

znation:
@hoytak, @seanses: how about we merge this PR, then open separate issues for (1) renaming libmagic, and (2) updating the file type list using the Wikipedia data (with some manual effort)?

seanses (approved):

hoytak:
Sounds good to me!

PR #60, opened 2023-07-19
Rewrite merge order of shards in session directory 

hoytak:
This PR changes the merge order of shards in the session directory to be linear in line with file creation.  
This both simplifies the algorithm and allows for the future insertion of shard hints based on dedup accesses.


rust/mdb_shard/src/merging.rs
```diff
@@ -33,134 +33,129 @@ fn write_shard(target_directory: &Path, data: &[u8]) -> Result<(MerkleHash, Path
...
+            // Put the new shard on the return stack and move to the next unmerged one
+            finished_shards.push(new_sfi);
+        }
+
+        // Delete the old ones.
+        for p in files_to_delete.iter() {
+            // In rare cases, there could be empty shards or shards with
+            // duplicate entries and we don't want to delete any shards
+            // we've already finished
+            if *p != finished_shards.last().unwrap().path {
```
seanses, line 147: Such rare case can happen in any order and may not be the last finished shard?

hoytak: True...   It would mean that the previous shard information would have been merged into the current one, so it would be okay to delete in theory, but this would mean the list returned would be invalid.  I'll change it to check all of them.

hoytak: Fixed.

PR #65, opened 2023-07-20
Inter shard reference list in shard file; hints for dedup

hoytak:
Got shard hint info added into shard files.  Shard files now maintain an optional list of other referenced shards that can be used for deduplication of a new version of any of the files.

Shard format is fully forward and backwards compatible.


rust/mdb_shard/src/intershard_reference_structs.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn num_bytes(&self) -> u64 {
+        (size_of::<IntershardReferenceSequenceHeader>()
+            + self.entries.len() * size_of::<IntershardReferenceSequenceEntry>()) as u64
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty()
+    }
+
+    pub fn merge(self, other: IntershardReferenceSequence) -> Self {
```
seanses, line 164: Is this function used anywhere?

rust/mdb_shard/src/intershard_reference_structs.rs
```diff
@@ -0,0 +1,279 @@
...
+        let mut entries = Vec::with_capacity(metadata.num_entries as usize);
+        for _ in 0..metadata.num_entries {
+            entries.push(IntershardReferenceSequenceEntry::deserialize(reader)?);
+        }
+
+        Ok(Self { metadata, entries })
+    }
+}
+
+pub fn write_out_with_new_intershard_reference_section<R: Read + Seek>(
```
seanses, line 224: And is this function used?

rust/mdb_shard/src/shard_file_manager.rs
```diff
@@ -210,17 +185,19 @@ impl ShardFileManager {
...
-            })
-            .await?
-        {
-            Ok(Some(si))
-        } else {
-            Ok(None)
+        for (si, is_permanent) in self.shard_file_lookup.read().await.values() {
+            debug!("Querying for hash {:?} in {:?}.", &query_hashes[0], si.path);
+            if let Some((count, fdse)) = si.chunk_hash_dedup_query(query_hashes)? {
+                if *is_permanent {
```
seanses, line 191: This doesn't track references to shards in the session directory. Is that the design choice? If we don't track I don't see the point to merge shards in the order of mtime?

rust/mdb_shard/src/intershard_reference_structs.rs
hoytak: Not in this PR; it will be in the dependent PR after this.

rust/mdb_shard/src/intershard_reference_structs.rs
hoytak: Again, not in this PR; it will be in the dependent PR after this.

rust/mdb_shard/src/intershard_reference_structs.rs
seanses: I see

rust/mdb_shard/src/shard_file_manager.rs
hoytak: That is correct.  Because of merging, we don't know the hashes for those shards, so we do this part when we're processing the session directory.  See  https://github.com/xetdata/xet-core/pull/66#issue-1816374906 -- this explains why it must be done in order.

seanses (approved):

PR #66, opened 2023-07-21
Merging session directory adds references to other shards as hints.  

hoytak:
This PR adds functionality to the shard session directory merge algorithm to add references to other shards within that session.  With this PR, it is generally true that the intershard reference hints in a shard containing the file reconstruction information for a given file will contain references to all shards that describe dedupable chunks referenced by that file.  

Building on https://github.com/xetdata/xet-core/pull/65, this PR adds in intershard hints that reference other shards in a given session.   It does this by building a table of CAS hash to shard hash for existing scanned shards, then loading all the file reconstruction information and looking at the CAS hashes referenced.  The shard hash for each referenced CAS hash is then added to the shard hints.  Note that this information is already present for all permanent shards as tracked at dedup time, but it is not present for any of the shards within a given session.   

Note that because the file reconstruction information will reference CAS hashes in any previously created shard, adding this information in must be done in order of creation (as https://github.com/xetdata/xet-core/pull/60 ensures).


seanses (commented):
This contains some code changes in #65, maybe merges in #65 first and rebase this PR or merges main so it's easy to see what changed.

PR #67, opened 2023-07-24
Fix git-xet --version

ylow:
With the gitxetcore seperation, the command parsing is in gitxetcore which affects the name of the program printed by git-xet -V making it print "gitxetcore" instead.
This forces the name to "git-xet" instead of "gitxetcore"


rajatarya (approved):
LGTM!

seanses (approved):

PR #69, opened 2023-07-25
Potential fix for wrong Python binary detection

ylow:
It is difficult to reliably identify the location of the python binary which was used to run the 'xet' command. This is required to "relaunch" xet as a daemon. Here, instead we directly relaunch the 'xet' command directly, hoping that the system will "do the right thing" (i.e. exec the script with the shebang command at the start of 'xet').

### DIFF
--- a/rust/gitxetcore/src/command/mount.rs
+++ b/rust/gitxetcore/src/command/mount.rs
@@ -341,20 +341,20 @@ If you use a git UI, point it to the raw path.
         )?;
     }
 
-    // use std::env::current_exe to find ourselves if we have it
+    // If invoked from Python, we use argv[1]
+    // else use std::env::current_exe to find ourselves if we have it
     // otherwise run "git xet"
-    let mut command = if let Ok(curexe) = std::env::current_exe() {
-        let mut command = Command::new(curexe);
-        if args.invoked_from_python {
-            let argv: Vec<String> = std::env::args().collect();
-            if argv.len() < 2 {
-                return Err(errors::GitXetRepoError::Other(
-                    "Unable to find python script to invoke".into(),
-                ));
-            }
-            command.arg(&argv[1]);
+    let mut command = if args.invoked_from_python {
+        // we will shell exec the script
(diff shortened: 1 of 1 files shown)


ylow:
Unclear if this will work correctly on windows.

hoytak:
Wouldn't it be more reliable to always launch xet with PYTHON_EXECUTABLE set to `sys.executable` and just query that?

ylow:
Not sure I understand. What are you proposing? xet is launched from the command line so I don't have control over the environment variables. And sys.executable is not necessarily right either (docs are unclear if it resolves correctly if `python` is a symlink. In this case I want the original symlink, and not the final real path).

ylow:
Actually. I think this will actually work on windows. The way windows python does the "xet" cli is to well... it makes an executable.

hoytak:
I am fine if this works.  I am quite sure sys.executable resolves to the original symlink; this is actually how virtual envs function.   (I happen to know due to ... reasons ... that if sys.executable does not resolve to the symlink, then python virtual envs break).

hoytak (approved):
Okay by me if this works.  I think the correct logic is still to use sys.executable, but won't hold things up for that.

ylow:
I see. Maybe let me adjust the logic and switch it to pass sys.executable instead. That annoying will require me to make the fix in 2 repositories.

port8080:
I know very little abou these tradeoffs working with windows and virtual env. My ack is a no-op here.

ylow:
Ok. Switched to use a passed python executable. Which is likely the least intrusive fix as we know that re-execing Python does already work on windows. If sys.executable is correct, this should just work *better* and we avoid needing to test against windows oddities. Note this will not compile with pyxet until a matching pyxet PR goes in.

hoytak:
Could also call directly, or repro this logic?  
https://github.com/python/cpython/blob/260ec2c36abd73bac51489108409160427979ede/Modules/getpath.c#L526-L614

ylow:
Will rather not reproduce the logic. I have pyxet pass sys.executable via cli args.

PR #72, opened 2023-07-25
MDBv2 Mount optimizations

ylow:
This reduces llama2 mount time from a couple of minutes on my home internet to 3.5s.

 - Do not download shards on MDBv2 repos.
 - Cache reconstruction lookups
 - some documentation
 - removed a misleading printout with number of objects / bytes (since the NFS is now lazily constructed... we do not actually have this number.)
Note that further optimizations like directly using the xetblob interface rather than the git methods, can reduce this to instant.

### DIFF
--- a/rust/gitxetcore/src/smudge_query_interface.rs
+++ b/rust/gitxetcore/src/smudge_query_interface.rs
@@ -14,7 +14,7 @@ use tracing::info;
 use crate::constants::FILE_RECONSTRUCTION_CACHE_SIZE;
 use crate::data_processing_v2::GIT_XET_VERION;
 use shard_client::GrpcShardClient;
-use tokio::sync::Mutex;
+use std::sync::Mutex;
 
 #[derive(PartialEq, Default, Clone, Debug, Copy)]
 pub enum SmudgeQueryPolicy {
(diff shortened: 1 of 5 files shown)


rust/gitxetcore/src/smudge_query_interface.rs
```diff
@@ -52,6 +55,8 @@ pub struct FileReconstructionInterface {
     pub smudge_query_policy: SmudgeQueryPolicy,
     pub shard_manager: Arc<ShardFileManager>,
     pub shard_client: Option<GrpcShardClient>,
+    pub reconstruction_cache:
+        Mutex<LruCache<merklehash::MerkleHash, (MDBFileInfo, Option<MerkleHash>)>>,
```
jgodlew, line 59: nit: we can get better performance by using a `std::sync::Mutex` instead of a `tokio::sync::Mutex` since we are not holding the lock across await points ([recommended in tokio docs](https://tokio.rs/tokio/tutorial/shared-state#use-tokios-asynchronous-mutex)).

jgodlew (approved):
LGTM! Just a small comment/FYI that we don't need the overhead of a tokio lock and can just use a sync lock instead.

:shipit:

seanses (approved):

rust/gitxetcore/src/smudge_query_interface.rs
ylow: Oh interesting. Good to know.

ylow: fixed

znation (approved):

PR #74, opened 2023-07-25
XetBlob now downloads shard hints on open_for_write

hoytak:
This PR completes the plumbing of the shard hints by causing open_for_write to first download any shards associated with a given file before beginning the write.  

There are usability tweaks that could be implemented in future PRs: 
- Only download shards for files above a certain size.
- Limiting the number of shards downloaded.
- Background downloading the shards as the write is happening.

Also, this needs thorough testing.


rust/gitxetcore/src/command/dir_summary.rs
```diff
@@ -62,12 +62,7 @@ pub async fn dir_summary_command(config: XetConfig, args: &DirSummaryArgs) -> er
     } else {
         tracing::info!("Recomputing");
         // recompute the dir summary
-        let summaries = compute_dir_summaries(
-            &repo,
-            &args.reference,
-            args.recursive,
-        )
-        .await?;
+        let summaries = compute_dir_summaries(&repo, &args.reference, args.recursive).await?;
```
ylow, line 65: You have a wierd reformatting thing again... Can you minimize this PR?

rust/xetblob/src/xet_repo.rs
```diff
@@ -334,6 +335,24 @@ impl XetRepoWriteTransaction {
...
+                .perform_stat_query(self.remote_base_url.clone(), &self.branch, filename)
+                .await?
+            {
+                let ptr_file =
+                    PointerFile::init_from_string(&String::from_utf8_lossy(&body), filename);
+
+                if ptr_file.is_valid() {
+                    // TODO: strategies to limit this, and limit the number of shards downloaded?
+                    let file_hash = MerkleHash::from_hex(ptr_file.hash())?;
+                    tr_v2.fetch_hinted_shards_for_file(&file_hash).await?;
```
ylow, line 351: Lets not do this right here. I worry that this will introducing a ton of overhead for large recursive uploads. Can we add this as a seperate method instead and plumb it through pyxet? It is easier to introduce more elaborate heuristics on the Python side. I am thinking:
 - user hinting
 - for large cps, we will do this.
 - for recursive cps we can poll the folder and fetch shards of large files in the same folder.
 - For xetfs file writes we could buffer X MB of writes in Python, then look for hints.

Or stuff like that.

ylow (approved):

ylow (changes requested):
see comments.

rust/xetblob/src/xet_repo.rs
```diff
@@ -328,6 +332,93 @@ impl XetRepo {
             shard_session_dir,
         })
     }
+
+    // Fetches all the shard in the hints corresponding to one or more source endpoints.
+    // Endpoints are specified by a list of (branch, path) tuples.
+    pub async fn fetch_shards_for_dedup<S: AsRef<String>>(
```
ylow, line 338: endpoints ambiguous. Just "files"?
What is min_dedup_chunk_count for?

ylow (approved):

PR #78, opened 2023-07-27
Temporary forward to get xethub to compile.

hoytak:
Turns out dependent packages depend on shard_file, which got renamed to shard_format in a previous commit.  This PR adds in a temporary shard_file.rs with forward imports from shard_format to get dependencies to compile until that changes.


seanses:
If this is for xethub can we directly update xethub code?

hoytak:
We could directly update xethub, true.  Let me do that instead.

PR #80, opened 2023-07-27
Pin Intaglio version

hoytak:
Pins intaglio version to <1.9.0 and also reverts 19c0d6e624605e478a35a906bed977514fcbc819, which adds cargo.lock


seanses (approved):

lopopolo:
Hi folks, I maintain `intaglio`.  1.9.0 fixes an unsoundness issue. Is there anything I missed in packaging this release that makes y'all unable to adopt it? It looks like y'all build with Rust 1.70.0 in CI. The only change that I expected to cause disruption in this release was the MSRV bump to 1.58.0.

Let me know if I can help at all and I hope y'all are getting value out of the crate.

hoytak:
> Hi folks, I maintain `intaglio`. 1.9.0 fixes an unsoundness issue. Is there anything I missed in packaging this release that makes y'all unable to adopt it? It looks like y'all build with Rust 1.70.0 in CI. The only change that I expected to cause disruption in this release was the MSRV bump to 1.58.0, which is in line with the crate's documented policy of MSRV bumps being acceptable in minor releases.
> 
> Let me know if I can help at all and I hope y'all are getting value out of the crate.

Thank you @lopopolo!  Yes, intaglio is quite useful, and thanks for being proactive here!   The issue we hit is the following: 

```
325 | impl NFSFileSystem for MirrorFS {
    |                        ^^^^^^^^ `NonNull<OsStr>` cannot be sent between threads safely
    |
    = help: within `intaglio::internal::Interned<OsStr>`, the trait `Send` is not implemented for `NonNull<OsStr>`
```

and it seems that this was introduced in 1.9.0.  I will file an issue on your end as well.

lopopolo:
Thank you. This is a regression on my end. The symbol tables are no longer `Send`, `Sync`, and `Unpin` due to the way I fixed the soundness hole. I'm missing some trait impls.

Later today I'll yank 1.9.0 and release 1.9.1 with the fixes.

Apologies for the breakage and thanks for the report!

lopopolo:
This pin should no longer be necessary. I've yanked v1.9.0 which contained unexpected API breakage and re-released the fix for the soundness hole as v1.9.1. Thanks again folks!

PR #81, opened 2023-07-28
add parquete, jsonl, and zstd to list of file types

Saba9:
Correctly identify parquet, json, and zest files.


ylow (approved):

rust/libmagic/src/file_types.rs
```diff
@@ -82,6 +82,7 @@ static FILE_TYPES: phf::Map<&'static str, FileTypeInfo> = phf_map! {
     "dxf" => FileTypeInfo { friendly_type: "AutoCAD DXF", mime_type: "image/vnd.dxf"},
     "dwf" => FileTypeInfo { friendly_type: "Autodesk Design Web Format (DWF)", mime_type: "model/vnd.dwf"},
     "par" => FileTypeInfo { friendly_type: "BAS Partitur Format", mime_type: "text/plain-bas"},
+    "parquete" => FileTypeInfo { friendly_type: "Apache Parquete ", mime_type: "application/octet-stream"},
```
znation, line 85: `parquet` without an e at the end. And "Apache Parquet" without an e or space at the end.

znation (changes requested):

rust/libmagic/src/file_types.rs
Saba9: good catch

znation (approved):

PR #86, opened 2023-08-01
Added deduplication against current chunks in the CAS aggregator.

hoytak:
With this change, deduplication also happens against chunks stored in the common CAS aggregator, fixing the issue where many small files wouldn't see much deduplication.

### DIFF
--- a/rust/gitxetcore/src/data_processing_v2.rs
+++ b/rust/gitxetcore/src/data_processing_v2.rs
@@ -24,8 +24,8 @@ use pointer_file::PointerFile;
 use progress_reporting::DataProgressReporter;
 use std::mem::take;
 use tokio::sync::mpsc::Sender;
-use tokio::sync::watch;
 use tokio::sync::Mutex;
+use tokio::sync::{watch, RwLock};
 use tracing::{debug, error, info, info_span, warn};
 use tracing_futures::Instrument;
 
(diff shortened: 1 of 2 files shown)


hoytak:
Addresses https://github.com/xetdata/xethub/issues/3515

rust/gitxetcore/src/data_processing_v2.rs
```diff
@@ -45,10 +45,19 @@ use crate::summaries_plumb::WholeRepoSummary;
...
+// duplicate files.
+const CHUNK_IN_COMMON_AGGREGATOR: MerkleHash = MerkleHash::from_const([0, 0, 0, 1]);
+
 #[derive(Default)]
 struct CASDataAggregator {
     data: Vec<u8>,
     chunks: Vec<(MerkleHash, (usize, usize))>,
+
+    // A hash map of the chunks currently in the aggregator, along with their indices.
+    chunk_hash_map: HashMap<MerkleHash, usize>,
```
seanses, line 59: The function local aggregator also relies on a similar data structure `current_cas_block_hashes`. Since `CASDataAggregator` already incorporates a hash-> indices map, it might be better to also modify that part.

rust/gitxetcore/src/data_processing_v2.rs
```diff
@@ -409,6 +418,19 @@ impl PointerFileTranslatorV2 {
...
+
+                            file_info.push(FileDataSequenceEntry::new(
+                                CHUNK_IN_COMMON_AGGREGATOR,
+                                n_bytes,
+                                data_lb,
+                                data_ub,
+                            ));
+                            add_new_data = false;
                         } else if !file_info.is_empty()
                             && file_info.last().unwrap().cas_hash == MerkleHash::default()
```
seanses, line 435: And I think a similar optimization logic should be applied to `CHUNK_IN_COMMON_AGGREGATOR`

seanses: Hmm. So we will have this optimization in three places: against a function local aggregator, against a struct global aggregator, and against a permanent shard. Maybe some refactoring in the future.

PR #88, opened 2023-08-03
lower some debug to trace

ylow:


hoytak:
Weird.  Must have pushed an intermediate commit for this instead of the locally correct one.  Should work now.

hoytak (approved):

PR #91, opened 2023-08-09
begin_write_transaction now lets the synchronize step be optional.

hoytak:
While we're still figuring this out, the "Synchronize with remote" step is now enabled only if the environment variable XET_FETCH_ALL_SHARDS is set.  If this is not set, then this step is skipped, and all dedup relies on the shard hints mechanism.  However, this may have unforeseen consequences, so for now it may still be enabled for debugging and profiling.

### DIFF
--- a/rust/xetblob/src/xet_repo.rs
+++ b/rust/xetblob/src/xet_repo.rs
@@ -292,14 +292,16 @@ impl XetRepo {
             .map_err(|_| anyhow::anyhow!("Branch does not exist"));
 
         // Let's download all shards for now, delete this when shard hint lists work.
-        self.pull().await?;
-        sync_mdb_shards_from_git(
-            &self.config,
-            &self.config.merkledb_v2_cache,
-            GIT_NOTES_MERKLEDB_V2_REF_NAME,
-            true,
-        )
-        .await?;
+        if std::env::var_os("XET_FETCH_ALL_SHARDS").is_some() {
+            self.pull().await?;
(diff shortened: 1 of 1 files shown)


ylow:
I think the fix should simply be we do not fetch the shards for mdbv2.

ylow:
The config variable is fine, but we do need to pull for for mdbv1.

rust/xetblob/src/xet_repo.rs
```diff
@@ -291,22 +291,35 @@ impl XetRepo {
...
+            fetch_shards = true;
+        }
+
+        if let Some(v) = std::env::var_os("XET_FETCH_ALL_SHARDS") {
+            fetch_shards = v != "0";
+        }
+
+        if fetch_shards {
+            self.pull().await?;
+            sync_mdb_shards_from_git(
```
seanses, line 314: This function only downloads shards. The above call to `self.pull()` pulls in `refs/notes/merkledb*` and merges into v1 MDB if the repo is v1. So we don't call this function but need to call `self.pull()` if `PFTRouter::V1(_) = &self.translator.pft`.

hoytak: I think the above logic calls both for MDB V1?

hoytak: Ah, I see what you mean.

seanses (approved):

PR #94, opened 2023-08-16
unset readonly flag

seanses:
`std::fs::remove_file` currently corresponds to the `unlink` function on Unix and the `DeleteFile` function on Windows. `unlink` will succeed as long as having write access to the directory. `DeleteFile` fails if the file is read-only.
This PR unsets the read-only attribute for staging files before deletion.

Fix https://github.com/xetdata/xethub/issues/3891


ylow:
Do we know how it ends up being created as readonly?

seanses:
> Do we know how it ends up being created as readonly?

It's by design in the local client: https://github.com/xetdata/xet-core/blob/8c810c646027faea7f94414d9bcc5cb1b95205ed/rust/cas_client/src/local_client.rs#L310

ylow (approved):

PR #95, opened 2023-08-16
Fix: environment variables starting with XET_* corrupt configs.  

hoytak:
This fixes an issue in which XET_* environment variables are dumped into the serialized config variables, however they can't be parsed by the cfg loader on load, causing the config loading to be incorrect.  This fixes this issue.


jgodlew (approved):
LGTM! Thanks for looking into this and fixing it! :shipit:

PR #96, opened 2023-08-17
Possible fix for https://github.com/xetdata/xethub/issues/3890.

hoytak:
I'm not sure why the previous was failing, as the given pattern would have matched the regex in question.  However, this relies on the underlying file path library, so it should be more robust on windows.


ylow:
The rest of the function has a few unwraps should we just return None perhaps?

seanses (approved):
I confirm this fixes the "Bad file name format" error.

hoytak:
> The rest of the function has a few unwraps should we just return None perhaps?

The fact that the regex matched guarantees those to work; the field hash will be present and it will be 64 hex characters for the from_hex function.

PR #98, opened 2023-08-23
background puts support for CASClient, implemented in PassthroughStagingClient

ylow:
This PR changes a fundamental interface component of the CAS Client. It allows the put() method to be backgrounded. The background put() capability has been implemented in the PassthroughStagingClient (at the moment that is the most convenient location to enable the xetblob and hence Python API to get parallel uploads).

This means that for corectness, all puts() must be followed by a flush(). I fixed all the locations I can find.

### DIFF
--- a/rust/cas_client/src/caching_client.rs
+++ b/rust/cas_client/src/caching_client.rs
@@ -88,6 +88,11 @@ impl<T: Client + Debug + Sync + Send> Client for CachingClient<T> {
         self.client.put(prefix, hash, data, chunk_boundaries).await
     }
 
+    async fn flush(&self) -> Result<(), CasClientError> {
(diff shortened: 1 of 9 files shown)


ylow:
@jgodlew will love your eyes on this. I am using FuturesUnordered :-)

jgodlew (approved):
LGTM! :shipit:

jgodlew:
> I am using FuturesUnordered

I think this should be ok to use with `FuturesUnordered` since we don't have a particular ordering that could fill up the underlying futures queue (unlike `FuturesOrdered` in the deadlock situation). Also, the fact that the `client.put` method doesn't have a singleflight or other lock to prevent other tasks from running further reduces the chance of a deadlock.

PR #106, opened 2023-09-01
Faster xetblob download to disk

seanses:
Adds a `get` function to download an entire file to disk.
Adds a file open options to disable caching when downloading an entire file.


port8080 (approved):
:shipit: 📈

PR #114, opened 2023-09-11
Update to shard sync protocol.

hoytak:
Adds in an optional force field to the shard sync request to allow forced updates of the database.


port8080 (approved):
lgtm :shipit:

PR #117, opened 2023-09-11
Smudge now passes bad file reconstruction lookups through as pointers.

hoytak:
Previously, a bad file hash lookup caused an unrecoverable error.  Now, when looking up file reconstruction information fails, an error is printed but the original pointer file is passed through instead.


seanses (commented):
This deals with failed file reconstruction info lookup, what about bad (corrupted) file reconstruction info? Do we want to also include that case in this PR?

hoytak:
> This deals with failed file reconstruction info lookup, what about bad (corrupted) file reconstruction info? Do we want to also include that case in this PR?

I worry that it would be difficult to do this without swallowing possibly useful errors needed for other debugging cases.  But I'm open to changing that. Currently, I'd like to be conservative here about changing the behavior.

PR #120, opened 2023-09-13
Update the current progress reporting class for pyxet use.

hoytak:
This PR updates the current progress bar used by the git_stream functions and CAS uploading routines to avoid &mut self, thus allowing it to be easily wrapped by pyxet and passed between python and rust.  This has the additional advantage of avoiding a mutex, as was the case in git_stream.  The base functionality and format has not changed.

To test out the different formats and use cases, run the produced executable test_progress_reporting.   See progress_reporting/src/main.rs for example code.  

```
Testing progress bar (no totals, no bytes): 4 completed, done.
Testing progress bar (total count, no bytes): 100% (5 / 5), done.
Testing progress bar (no totals, only bytes): 14.65 KiB | 9.69 KiB/s, done.
Testing progress bar (only bytes + total): 100% (19.53 KiB / 19.53 KiB) | 9.70 KiB/s, done.
Testing progress bar (no totals, both count + bytes): 27 / 27, 14.65 KiB | 9.69 KiB/s, done.
Testing progress bar (Total count): 100% (30 / 30), 14.65 KiB | 9.69 KiB/s, done.
Testing progress bar (Total count + total bytes): 100% (30 / 30), 19.53 KiB | 9.68 KiB/s, done.
```

### DIFF
--- a/rust/cas_client/src/staging_client.rs
+++ b/rust/cas_client/src/staging_client.rs
@@ -116,10 +116,10 @@ impl StagingUpload for StagingClient {
         );
 
         let pb = if self.progressbar && !entries.is_empty() {
-            let mut pb =
-                DataProgressReporter::new("Xet: Uploading data blocks", Some(entries.len()));
+            let pb =
+                DataProgressReporter::new("Xet: Uploading data blocks", Some(entries.len()), None);
 
-            pb.register_progress(Some(0), 0); // draw the bar immediately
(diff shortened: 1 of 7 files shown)


seanses (approved):
LGTM! @erinys may want to double check on the style.

port8080 (approved):
🚢 :shipit:

PR #125, opened 2023-09-16
git xet smudge -i <hash> now materializes a file, independent of repo 

hoytak:
This PR: 
- Adds checks in data_processing, shard manager, and other places to allow limited operation without a repo present. 
- Added --id/-i flag to git-xet smudge to take a raw hash instead of a pointer file and smudge based on that.  Thus:

```
git xet smudge -i <hash> -o <output_file> 
```

will materialize the file given by the hash ID to the output_file given.

### DIFF
--- a/rust/gitxetcore/src/command/smudge.rs
+++ b/rust/gitxetcore/src/command/smudge.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use clap::Args;
+use merklehash::MerkleHash;
 
 use crate::async_file_iterator::AsyncFileIterator;
 use crate::config::XetConfig;
@@ -24,14 +25,20 @@ use thiserror::Error;
 /// git xet pointer -f input | git xet --cas smudge > output
 /// # input and output should be the same file
 /// ```
+///
 pub struct SmudgeArgs {
     /// If passthrough is set (default is disabled),
     /// we will also "smudge" non-pointer files, by passing it through
     /// direct to stdout.
     #[clap(short, long)]
     passthrough: bool,
+
     #[clap(long, short)]
     filename: Option<PathBuf>,
+
+    /// Lookup file by hash id
+    #[clap(long, short)]
+    id: Option<String>,
 
     #[clap(long, short)]
     output: Option<PathBuf>,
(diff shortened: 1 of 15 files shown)


assafvayner:
Tested this out locally and managed to hit an interesting bug.
- using release build git-xet off this branch
- hash used can be find in this [pointer file on github](https://github.com/xetdata/assaf-test3/blob/main/test) (ac6f3356c4ef9ac324c0629c88c31cca202553981528078e4de6ab310550abf0)
  - the file that this hash relates to is in prod CAS.
 
 
- running `git xet smudge -i <hash>` inside of a github repo shows `Hash not found`
- running the same command outside the repo succeeds in getting the data

Screenshot 2023-09-15 at 9 12 30 PM

hoytak:
Interesting.  It works for me in a couple smoke tests... 

It should attempt to read the merkledb from the local repo, and then, if it's not present there, query the shard server.   Could you send me the run with XET_LOG_LEVEL=debug?  I'm curious how it's trying to work.   

The one thing that I'm wondering about here is whether the repo changed the config, particularly the CAS endpoint or query policy (In a config with a local:// CAS endpoint, it doesn't query the server).  This could cause the issue.

seanses:
Looks like it's because that when `git-xet` is running under `~/xetworkspace/xet-core`, it picks up the information of this repo and sees it as MDBv1 (didn't detect a special note under this repo) and tries to query against the merkle tree:
```
2023-09-16T04:19:56.295420Z DEBUG gitxet: merkledb/src/merklememdb.rs:78: Opening DB "/Users/assafvayner/xetworkspace/xet-core/.git/xet/merkledb.db" command="smudge"
```.

But I'm not sure why `git-xet` detects MDBv2 when it's running under `~/xetworkspace`. What do you have under this directory? @assafvayner

assafvayner:
My ~/xetworkspace has multiple standard github repositories cloned. couple other folders, but no .git directory, it's not a git repo. The thing I hope for is if we provide the hash, and error at any point reading a merkledb v1 or v2 that we can presume v2 and proceed talking with the shard server since we don't need the merkledb locally existing before the reconstruction operation.

seanses:
Traced down the cause: when a git repo is not detected it assumes a v2 smudge path, so when you run `git-xet` under `~/xetworkspace` it defaults to v2 and smudged successfully. I think the behavior is expected: if you run `git-xet` under a repo, it should respect the information of that repo.

assafvayner (approved):

seanses (approved):

PR #119, opened 2023-09-12
github actions

assafvayner:
Use github actions over circleci and run on windows, mac and linux as opposed to just linux.
Fix tests failures on Windows.

Resolves https://github.com/xetdata/xethub/issues/4157, https://github.com/xetdata/xet-core/issues/163

PR #126, opened 2023-09-18
fail at failing to fill buffer

assafvayner:


assafvayner:
included in assaf/github-actions #119

PR #127, opened 2023-09-19
Fix for rename issue

port8080:
Context: https://docs.gitlab.com/ee/user/project/repository/index.html#what-happens-when-a-repository-path-changes


port8080:
Fixes https://github.com/xetdata/xethub/issues/3899

The `is_repo` call was using the existence of branches on a repo as the test. In the case of repo renames, it was following redirects which caused it to return true when the repo had been renamed.

Testing done:

- Built a local pyxet using maturin with this change
- Confirmed that these two commands worked back-to-back
```
xet repo rename xet://port8080/Repo xet://port8080/Repo-edited
xet repo rename xet://port8080/Repo-edited xet://port8080/Repo
```

ylow (approved):

PR #129, opened 2023-09-20
error codes for GitXetRepoError

assafvayner:
This PR makes git-xet return different error codes for different error types. This is needed for better error cause parsing in internal projects that use git-xet.


ylow (approved):

hoytak (approved):

hoytak:
Great!

PR #131, opened 2023-09-22
Update version to 0.11.5

hoytak:
Post Release: Updates version to 0.11.5.


assafvayner (approved):
Need to add "" surrounding version in tomls

PR #133, opened 2023-09-27
Fix so that merkledb version determination is correct.

hoytak:
Currently, the errors in the merkledb logic mean that in some cases, the merkledb version is incorrectly detected.  When this happens, some of the smudging and initialization logic has gotten messed up.  This PR addresses this and allows repositories without merkledb notes to be properly initialized.


rust/gitxetcore/src/errors.rs
```diff
@@ -98,6 +98,12 @@ pub enum GitXetRepoError {
 
     #[error("Authentication Error: {0}")]
     AuthError(anyhow::Error),
+
+    #[error("Xet Repo operation attempted before repo is initialized : {0}")]
+    RepoUnitialized(String),
```
seanses, line 103: Typo, `Uninitialized`

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -1582,7 +1661,7 @@ impl GitRepo {
         }
 
         // Only need to install guard notes when this is an upgrade.
-        if self.mdb_version < *version {
+        if self.mdb_version < *version && self.mdb_version != ShardVersion::Unitialized {
```
seanses, line 1664: Why skip setting guard notes if it's uninitialized? Although this is not an issue when no one uses old clients.

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -513,6 +554,24 @@ impl GitRepo {
...
+    pub fn install_gitxet_from_filter_process(&self) -> Result<()> {
+        // Initialize a local version of the
+        info!(
+            "Configure Merkle DB and repo salt associated with repo {:?}",
+            &self.repo_dir
+        );
+
+        self.set_repo_mdb_to_v2_from_uninitialized()?;
+
+        // self.set_repo_salt()?;
```
seanses, line 570: Why do we skip setting repo salt?

rust/gitxetcore/src/merkledb_shard_plumb.rs
```diff
@@ -738,6 +720,49 @@ pub fn write_mdb_version_guard_note(
...
+    if !repo_path.exists() {
+        info!("get_mdb_version: Repo path {repo_path:?} does not exist; defaulting to ShardVersion::V2.");
+        return Ok(ShardVersion::get_max());
+    }
+
+    let Ok(repo) = open_libgit2_repo(Some(repo_path)).map_err(|e| {
+        info!("get_mdb_version: Repo path {repo_path:?} does note appear to be a repository ({e}); defaulting to ShardVersion::V2.");
+        e
+    }) else {
+        return Ok(ShardVersion::V2);
```
seanses, line 733: Also this should be `Uninitialized`.

rust/mdb_shard/src/shard_version.rs
```diff
@@ -8,8 +8,12 @@ pub const MDB_SHARD_FOOTER_VERSION: u64 = MDB_SHARD_VERSION;
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Copy, Clone, Debug, Default)]
 pub enum ShardVersion {
+    /// In a git repo that does not have merkledb v1 or v2 elements.
+    Unitialized = 0,
```
seanses, line 12: Typo, `Uninitialized`

rust/gitxetcore/src/git_integration/git_repo.rs
assafvayner: Also delete the comment if it's truly not needed. The behavior regarding the salt should be write it if it's not found correct?

rust/gitxetcore/src/git_integration/git_repo.rs
hoytak: In this case, this function gets called from many different places.  Given that this functionality is only needed when cloning from a github repo (currently), I want to isolate it to that case.  
When first creating a repo, or when running git in a non-xet repo, the notes may not be available yet and I don't want there to be unintended side effects when running e.g. status or other commands.  At least with the filter, we can be confident that it's only being run in a xet repo.

rust/gitxetcore/src/git_integration/git_repo.rs
hoytak: This here was left over from debugging the previous command, where it was actually running the filter before the notes were fetched and thus causing the repo salts to end up conflicting.  Now we could call this, as the salt has been fetched, but in the github case it's been set remotely.

rust/gitxetcore/src/git_integration/git_repo.rs
seanses: I see, so we postpone setting the guard from init to the filter process.

seanses (approved):
Looks good, there is one typo in `GitXetRepoError:: RepoUnitialized`, should be `RepoUninitialized`

PR #135, opened 2023-09-29
Lazy clone and granular smudge

seanses:
Resolves https://github.com/xetdata/xethub/issues/4143.

To lazy clone a repo use option "--lazy", e.g `git xet clone [url] --lazy`. This will clone the repo with no pointer files smudged and create a file `.git/xet/lazyconfig` with one line of content: "pointer *", which means to keep all pointer files un-materialized. 

Users can add rules to this file, following syntax `[strategy] [target]`, like below:
`[pointer | p] [path]`: to keep pointer file `path` or files under directory `path` un-materialized.
`[smudge | s] [path]`: to materialize pointer file `path` or files under directory `path`.

When two rules can be applied to a common set of files, the rule with the longest matching target path will finally be applied. The scope of a rule is repo wide.

Use command `git xet lazy apply` to apply config changes to the working directory, or use `git checkout [reference]` to checkout files of a different branch / commit.

Use command `git xet lazy check` to help check the correctness of the config file, specifically, if the rules have correct syntax, and if there are conflicting rules (rules with the same target path but different strategy).

Use command `git xet lazy match [path]` to help debug the config. It will find the matched rule to the given path and print it to the terminal.

### DIFF
--- a/rust/gitxetcore/src/command/lazy.rs
+++ b/rust/gitxetcore/src/command/lazy.rs
@@ -14,7 +14,7 @@ enum LazyCommand {
     /// rules are non-conflicting.
     Check,
     Match(LazyMatchArgs),
-    /// After editting a config file, apply changes to the
+    /// After editing a config file, apply changes to the
     /// working directory.
     Apply,
 }
@@ -89,7 +89,7 @@ async fn lazy_apply_command(cfg: &XetConfig) -> Result<()> {
 
     if !repo.repo_is_clean()? {
         return Err(GitXetRepoError::InvalidOperation(
-            "repo is dirty, abort...".to_owned(),
+            "Repo is dirty; commit your changes and try this operation again.".to_owned(),
         ));
     }
 
--- a/rust/gitxetcore/src/command/lazy.rs
+++ b/rust/gitxetcore/src/command/lazy.rs
@@ -59,9 +59,9 @@ async fn lazy_check_command(cfg: &XetConfig) -> Result<()> {
 
         Ok(())
     } else {
-        Err(GitXetRepoError::InvalidOperation(format!(
-            "lazy config file doesn't exist"
-        )))
+        Err(GitXetRepoError::InvalidOperation(
+            "lazy config file doesn't exist".to_owned(),
+        ))
     }
 }
 
@@ -75,9 +75,9 @@ async fn lazy_match_command(cfg: &XetConfig, args: &LazyMatchArgs) -> Result<()>
 
         Ok(())
     } else {
-        Err(GitXetRepoError::InvalidOperation(format!(
-            "lazy config file doesn't exist"
-        )))
+        Err(GitXetRepoError::InvalidOperation(
+            "lazy config file doesn't exist".to_owned(),
+        ))
     }
 }
 
@@ -88,9 +88,9 @@ async fn lazy_apply_command(cfg: &XetConfig) -> Result<()> {
     let repo = GitRepo::open(cfg.clone())?;
 
     if !repo.repo_is_clean()? {
-        return Err(GitXetRepoError::InvalidOperation(format!(
-            "repo is dirty, abort..."
-        )));
+        return Err(GitXetRepoError::InvalidOperation(
+            "repo is dirty, abort...".to_owned(),
+        ));
     }
 
     std::fs::remove_file(cfg.repo_path()?.join("index"))?;
--- a/rust/gitxetcore/src/command/lazy.rs
+++ b/rust/gitxetcore/src/command/lazy.rs
@@ -1,0 +1,101 @@
+use clap::{Args, Subcommand};
+use lazy::lazy_config::LazyConfig;
+use std::path::PathBuf;
+
+use crate::config::XetConfig;
+use crate::errors::{GitXetRepoError, Result};
+use crate::git_integration::git_repo::GitRepo;
+use crate::git_integration::git_wrap;
+
+#[non_exhaustive]
+#[derive(Subcommand, Debug)]
+enum LazyCommand {
+    /// Checks if a config file has correct syntax and
+    /// rules are non-conflicting.
+    Check,
+    Match(LazyMatchArgs),
+    /// After editting a config file, apply changes to the
+    /// working directory.
+    Apply,
+}
+
+/// Given a path relative to the repository root, find
+/// the rule that will be applied.
+#[derive(Args, Debug)]
+struct LazyMatchArgs {
+    path: PathBuf,
+}
+
+// THIS "SHIM" STRUCT IS MANDATORY
+#[derive(Args, Debug)]
+pub struct LazyCommandShim {
+    #[clap(subcommand)]
+    subcommond: LazyCommand,
+}
+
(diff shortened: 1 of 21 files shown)


rust/gitxetcore/src/command/lazy.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::git_integration::git_wrap;
+
+#[non_exhaustive]
+#[derive(Subcommand, Debug)]
+enum LazyCommand {
+    /// Checks if a config file has correct syntax and
+    /// rules are non-conflicting.
+    Check,
+    Match(LazyMatchArgs),
+    /// After editting a config file, apply changes to the
```
hoytak, line 17: nit: comment typo

rust/gitxetcore/src/command/lazy.rs
```diff
@@ -0,0 +1,101 @@
...
+
+async fn lazy_apply_command(cfg: &XetConfig) -> Result<()> {
+    // rm .git/index
+    // git checkout -f
+
+    let repo = GitRepo::open(cfg.clone())?;
+
+    if !repo.repo_is_clean()? {
+        return Err(GitXetRepoError::InvalidOperation(
+            "repo is dirty, abort...".to_owned(),
```
hoytak, line 92: Use more actionable error message: 
"Repo is dirty; commit your changes and try this operation again."

rust/gitxetcore/src/command/lazy.rs
```diff
@@ -0,0 +1,101 @@
...
+        ));
+    }
+
+    std::fs::remove_file(cfg.repo_path()?.join("index"))?;
```
hoytak, line 96: I think this is <root>/index, not <root>/.git/index.

hoytak: Also, why is it necessary to remove the index file?   I think passing "HEAD" to the checkout command should accomplish this as well.

rust/lazy/src/lazy_config.rs
```diff
@@ -0,0 +1,163 @@
...
+            // comments
+            if line_content.trim().starts_with('#') {
+                continue;
+            }
+            let rule: LazyRule = line_content.parse()?;
+            rules.push(rule);
+        }
+
+        // sort the rules by alphabetical order, and wildcard always at the top
+        rules.sort_by(|a, b| {
```
hoytak, line 37: Why would we need to sort these rules?  

Are we able to handle things like:


'*' smudge
'a/*' pointer
'a/b/*'  smudge
```
Here, if my rule file is like this, I would expect e.g. a/data.dat to be a pointer, but a/b/foo.bar to be smudged.  I'm not sure sorting them would preserve the precedence.

File: rust/lazy/src/lazy_config.rs [hunk 5]
```diff
@@ -0,0 +1,163 @@
...
+pub struct LazyConfig {
+    rules: Vec<LazyRule>,
+}
+
+impl LazyConfig {
+    pub async fn load(reader: &mut impl BufRead) -> Result<Self> {
+        let mut rules: Vec<LazyRule> = vec![];
+
+        for line in reader.lines() {
+            let line_content = line?;
```
hoytak, line 23: let line_content = line?.trim();

rust/lazy/src/lazy_rule.rs
```diff
@@ -0,0 +1,142 @@
...
+pub struct LazyRule {
+    pub path: String,
+    pub strategy: LazyStrategy,
+}
+
+impl FromStr for LazyRule {
+    type Err = LazyError;
+
+    fn from_str(s: &str) -> Result<Self> {
+        let split = s.trim().split_once(' ').ok_or_else(|| {
```
hoytak, line 38: This won't support windows with the space here, or filenames containing \ in linux.  I would instead just do an rsplit on ' ' and then use everything to the left as the path.

hoytak, line 38: Instead of splitting on just space, I would use something that allows splitting on tabs as well: 
```
let s = s.trim();
let Some(n) = s.find(|c| c.is_whitespace()) else { return LazyError::InvalidRule(format!("Failed to parse line {s:?}")); };
let (rule, path) = s.split_at(n);
let path = path.trim(); 
```

rust/lazy/src/lazy_rule.rs
```diff
@@ -0,0 +1,142 @@
...
+            .trim_matches(std::path::MAIN_SEPARATOR)
+            .to_owned();
+
+        Ok(LazyRule { path, strategy })
+    }
+}
+
+impl LazyRule {
+    pub fn match_path_prefix(&self, path: impl AsRef<Path>) -> Result<bool> {
+        let path = path
```
hoytak, line 55: I am wondering if it would be better to use a prebuilt glob matching library for this, e.g. https://docs.rs/globset/latest/globset/.  This would give us more flexibility and perhaps be easier / more reliable to work with.

rust/lazy/src/lazy_config.rs
seanses: Wildcards are not supported at the moment, but only in the scenario that a user defines something like `pointer a/data*`, as he wants to keep files under `a/` with prefix "data" as pointer files.

The exact behavior you want is supported, but defined in a different way:

```
smudge *
a pointer
a/b smudge
```

The reason to sort the rules by target path is that I can find the longest match first to implement the above behavior.

rust/lazy/src/lazy_rule.rs
seanses: Good point, I can change to `split_whitespace`

rust/lazy/src/lazy_rule.rs
seanses: I thought about glob matching, but haven't figured out the precedence between the below two rules applying to file `foo/bar/data1`:
```
smudge foo/bar/data*
pointer f*/b*/data1
```
So for simplicity I didn't implement glob matching.

seanses: I think we can get more feedback from users and product during trial of this feature.

rust/gitxetcore/src/command/lazy.rs
hoytak: Nice!

rust/lazy/src/lazy_config.rs
hoytak: Ah, ok, gotcha.  Interesting.  (Also, apologies -- for some reason I didn't see that notion document.)  If this is the consensus behavior, I'm down with it.  But I'm not convinced about the longest match behavior.

rust/lazy/src/lazy_rule.rs
hoytak: Fair.  That makes sense.

hoytak:
Ok to merge.  For some reason it won't let me add a review.

hoytak (approved):

rust/lazy/src/lazy_rule.rs
seanses: Actually, `split_once(char::is_whitespace)`

PR #138, opened 2023-10-05
Perform version check on certain commands

hoytak:
This PR adds a version check around many different git xet operations.  
- If the release notes on any version between the latest released version and the current version contain the letters CRITICAL, the user will be asked to upgrade every 2 hours, with a urgent warning:

```
**CRITICAL:**
A new version of the Xet client tools, {version}, is available at https://github.com/xetdata/xet-tools/releases and contains a critical bug fix from your current version.  Upgrading to this release immediately is strongly recommended.
```

- Otherwise every 24 hours the user will see:

```
A new version of the Xet client tools, {version}, is available at https://github.com/xetdata/xet-tools/releases.
```

Checking is done by polling github.com/xetdata/xet-tools/releases.

### DIFF
--- a/rust/gitxetcore/src/upgrade_checks.rs
+++ b/rust/gitxetcore/src/upgrade_checks.rs
@@ -44,7 +44,7 @@ fn get_critical_text() -> String {
 }
 
 fn get_version_info_filename() -> PathBuf {
-    let version_check_filename = match std::env::var("XET_VERSION_CHECK_FILENAME") {
+    let version_check_filename = match std::env::var("XET_UPGRADE_CHECK_FILENAME") {
         Ok(v) => v,
         Err(_) => VERSION_CHECK_FILENAME_HOME.to_owned(),
     };
--- a/rust/gitxetcore/src/upgrade_checks.rs
+++ b/rust/gitxetcore/src/upgrade_checks.rs
@@ -4,7 +4,7 @@ use crate::constants::CURRENT_VERSION;
 use chrono::{DateTime, Utc};
 use reqwest::{self, Client};
 use serde::{Deserialize, Serialize};
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 use version_compare::{self, Cmp};
 
 const GITHUB_REPO_OWNER: &str = "xetdata";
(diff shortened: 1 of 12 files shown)


rust/gitxetcore/src/upgrade_checks.rs
```diff
@@ -0,0 +1,431 @@
...
+            if version_is_newer(&version_tag, local_version) {
+                info!("get_newer_client_release_versions: Release {version_tag} newer than {local_version}.");
+                debug!(
+                    "get_newer_client_release_versions: Version {version_tag}; Body = {:?}",
+                    &release.body
+                );
+
+                newer_releases.push((version_tag, release.body));
+            } else {
+                break;
```
seanses, line 124: Is it a valid assumption that returned releases are sorted in descending tag name?

rust/gitxetcore/src/upgrade_checks.rs
```diff
@@ -0,0 +1,431 @@
...
+            if notification_happened {
+                self.inform_time = Utc::now();
+                self.save();
+            }
+        }
+    }
+
+    pub fn query_age_in_seconds(&self) -> u64 {
+        self.query_time
+            .signed_duration_since(Utc::now())
```
seanses, line 196: This function `signed_duration_since` should return negative value as it is `before - now`. Then taking the max between this value and 0 doesn't seem as intended.

rust/gitxetcore/src/upgrade_checks.rs
```diff
@@ -0,0 +1,431 @@
...
+            .num_seconds()
+            .max(0) as u64
+    }
+
+    pub fn inform_age_in_seconds(&self) -> u64 {
+        self.inform_time
+            .signed_duration_since(Utc::now())
```
seanses, line 203: Same here?

rust/gitxetcore/src/upgrade_checks.rs
hoytak: Yes, it is.

rust/gitxetcore/src/upgrade_checks.rs
hoytak: OH.  You are quite correct.  This could explain some intermittent issues.

rust/gitxetcore/src/upgrade_checks.rs
hoytak: Great catch.

seanses (approved):
LG!

PR #139, opened 2023-10-09
print repo salt along with mdb version

seanses:
This PR adds an option "merkledb version --with-salt" to print repo salt, if present, along with MDB version. Required by xetea returning repo info.


hoytak (approved):
LGTM.

PR #140, opened 2023-10-09
xetblob removes git dependency

seanses:
Drops git dependency for v2 repos. First query xetea for a repo's MDB version and repo salt.

Passed pyxet/tests/cli_cp_test.py against local xetea.

Depends on https://github.com/xetdata/xet-core/pull/139 and https://github.com/xetdata/xethub/pull/4333

Resolves https://github.com/xetdata/xethub/issues/4132, https://github.com/xetdata/xethub/issues/3927


hoytak (approved):
Ideally, I think we should use more direct plumbing than testing whether a repo just has .xetblob in it, but this does get the job done.   LGTM.

PR #141, opened 2023-10-10
Bugfix: Salt duplication bugfix.

hoytak:
The path previously could create a duplicate salt as incoming refs were not properly synced with local notes.  A salt should never be implicitly created in a repo; this eliminates that path and ensures the internal note sync happens properly.


seanses (approved):
LG!

PR #142, opened 2023-10-10
Allow (limited) calling of async functions from non async code.

hoytak:
For limited use; needed for a specific point in git integration. 
May not work well with the futures stream library.

### DIFF
--- a/rust/parutils/src/parallel_utils.rs
+++ b/rust/parutils/src/parallel_utils.rs
@@ -155,6 +155,38 @@ where
     Ok(take(&mut obj))
 }
 
+/// Allow an async function to be run from a non-async routine.
+/// Note that this blocks the current thread.  Use carefully!!!
+/// Also known to have bad interactions with futures_streams; don't
+/// use in those paths.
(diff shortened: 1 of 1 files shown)


rust/parutils/src/parallel_utils.rs
```diff
@@ -197,4 +229,14 @@ mod parallel_tests {
...
         Ok(())
     }
+
+    async fn value() -> Result<u64, anyhow::Error> {
+        Ok(42)
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_single_async() {
+        let v = block_on_async_function(|| async move { value().await }).unwrap();
```
seanses, line 239: What is the intended usage of this function? If it allows calling an async function inside a non-async function can the test be made to show that?

hoytak: It needs the tokio runtime, so the test is async, but note there isn't an await on the block_on_async_function.

seanses (approved):

PR #143, opened 2023-10-10
On implicit init, sync merkledb stuff too.

hoytak:
Likely needed for some pull workflows.

### DIFF
--- a/rust/gitxetcore/src/git_integration/git_repo.rs
+++ b/rust/gitxetcore/src/git_integration/git_repo.rs
@@ -1,6 +1,7 @@
 use is_executable::IsExecutable;
 use mdb_shard::error::MDBShardError;
 use mdb_shard::shard_version::ShardVersion;
+use parutils::block_on_async_function;
 use std::collections::{HashMap, HashSet};
 use std::fs::create_dir_all;
 #[cfg(unix)]
@@ -25,7 +26,7 @@ use tracing::{debug, error, info, warn};
 use crate::constants::*;
 use crate::data_processing::PointerFileTranslator;
 use crate::errors::GitXetRepoError::{self};
-use crate::errors::Result;
+use crate::errors::{convert_parallel_error, Result};
 use crate::git_integration::git_wrap;
 use crate::merkledb_plumb::{
     self, check_merklememdb_is_empty, merge_merkledb_from_git, update_merkledb_to_git,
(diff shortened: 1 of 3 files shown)


rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -1326,6 +1329,35 @@ impl GitRepo {
         Ok(())
     }
 
+    pub fn sync_notes_to_dbs_for_implicit_init(&self) -> Result<()> {
+        info!("XET sync_notes_to_dbs.");
+
+        self.sync_note_refs_to_local("merkledb", GIT_NOTES_MERKLEDB_V1_REF_SUFFIX)?;
```
seanses, line 1335: The filter process should trigger all these steps following repo open?

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -307,7 +308,9 @@ impl GitRepo {
             for r in remotes {
                 s.sync_remote_to_notes(&r)?;
             }
-            s.sync_note_refs_to_local("reposalt", GIT_NOTES_REPO_SALT_REF_SUFFIX)?;
+            // Sync in the notes and mdb stuff as needed.
+            s.sync_notes_to_dbs_for_implicit_init()?;
```
seanses, line 312: I think we can call `sync_notes_to_dbs` instead which does what this function is intended for

hoytak: Turns out we cannot.  The reason is annoying.  That function is async, so we would have to call it through the previous block_on_async function.  However,  libgit2 repo is not send or sync, so it can't be used in the call to that function, which means we have to create a version without the libgit2 repo attribute.  But the merkledb v1 path needs that repo information, so we need to create a path that avoids these constraints.

seanses: Makes sense

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -1326,6 +1329,35 @@ impl GitRepo {
...
+                &xet_config,
+                &merkledb_v2_cache_dir,
+                GIT_NOTES_MERKLEDB_V2_REF_NAME,
+                true, // with Shard client we can disable this in the future
+            )
+            .await
+        })
+        .map_err(convert_parallel_error)?;
+
+        debug!("XET sync_notes_to_dbs_implicit: merging summaries");
```
seanses, line 1356: We can drop this message if not actually merging summaries.

seanses (approved):

PR #147, opened 2023-10-13
Local config file to enable github forks.

hoytak:
This PR enables github integration that works through  forks and pull requests using an additional file in .xet/config.toml.  This file that gives the url of the repository where Xet is initialized.   Using this, on init, the filter process probes that repository to fetch the notes for the salt and the merkledb, if they exist.   This file looks like:

```
[upstream]
origin_type = "github"
user_name = "<repo user name>"
repo_name = "<repo name>"
```

This file, if present on HEAD, is picked up by the configuration.


rust/gitxetcore/src/constants.rs
```diff
@@ -17,6 +17,8 @@ pub const GIT_NOTES_REPO_SALT_REF_NAME: &str = "refs/notes/xet/reposalt";
 
 pub const GIT_LAZY_CHECKOUT_CONFIG: &str = "xet/lazyconfig";
 
+pub const GIT_REPO_SPECIFIC_CONFIG: &str = ".xet/config.toml";
```
seanses, line 20: Let's add a comment here says that this file is checked-in (in the working directory).

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -308,10 +372,25 @@ impl GitRepo {
...
+            }
+
+            if let Some(upstream_repo) = &s.xet_config.upstream_xet_repo {
+                info!("GitRepo::open: Adding remote upstream url {upstream_repo:?} to remotes to scan.");
+
+                // Create the remote url using the example form of the origin repo.
+                for remote_url in s.get_xet_upstream_remote_urls(upstream_repo)? {
+                    let _ = s.sync_remote_to_notes_with_custom_destination(
+                    &remote_url,
+                    Some("_upstream_xet_repo_"),
```
seanses, line 387: Where is this alternative remote set?

hoytak: So the backup, if that .xet/config.toml file isn't set, is to have the users manually add a remote that we can pull the repo salt from. 

The _upstream_xet_repo_ is just the location where the refs from that repo will be dumped on fetch.  It's not an actually remote that set.

seanses: I see.

rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -230,6 +235,65 @@ impl GitRepo {
...
+                let (ssh_is_first, ssh_only) = 'b: {
+                    // Try "origin" first, then try the rest of the remotes in order to find a github remote.
+                    for r_name in ["origin"]
+                        .into_iter()
+                        .chain(self.repo.remotes()?.iter().flatten())
+                    {
+                        if let Ok(r) = self.repo.find_remote(r_name) {
+                            let url = r.url().unwrap_or_default();
+                            if url.contains("github") {
+                                break 'b (url.contains("git@github"), true);
```
seanses, line 269: Not sure if we really need the `ssh_only` flag. It's a guess anyway?

hoytak: If the authentication for the https is wrong, then it hangs asking for user input...

seanses: Ah, that would be annoying. Make sense to me!

PR #148, opened 2023-10-13
Correct wrong info statement in upgrade check.

hoytak:
No logic change, just had the info!("...") statements wrong in one section.


seanses (approved):
LG!

PR #151, opened 2023-10-18
disable command path not implemented

seanses:
No actual functionality change. Just disable command paths in `git xet merkledb` that are not implemented.


hoytak (approved):
LGTM for now; obviously need to fill out the todo's.

PR #154, opened 2023-10-19
Refactor and fix up dir summary

znation:
* Make it ignore cache entries from previous versions of dir-summary
  (establishes this commit as "version 1" so all previous entries will
  be ignored).
* Remove deprecated command line parameter.
* Refactor to make the JSON output much nicer to consume and remove
  unused fields.
* Sort file types by extension
* Add several more file types for common file types


hoytak (approved):
LG!

PR #156, opened 2023-10-19
Local shard server mock to enable testing integration scenarios

hoytak:
Adds a simple local shard server using the existing shard file manager to enable local testing of operations that involve multiple repos but require the shard server to function correctly.  Now, when the CAS is specified using local://<path>, a shard server mock is set up that wraps a ShardFileManager instance running on <path>/shards, with the registration operation simply moving the shard from the local CAS in <path> to <paths>/shards and registering it.


rust/gitxetcore/src/config/xet.rs
```diff
@@ -411,13 +411,7 @@ impl XetConfig {
         smudge_query_policy: Option<SmudgeQueryPolicy>,
     ) -> Result<Self, ConfigError> {
         // First, ensure that the cas is not local; if it is, we're working completely local
```
seanses, line 413: We can drop this comment.

seanses (approved):
LG!

PR #158, opened 2023-10-19
print login success on login success

ylow:


rust/gitxetcore/src/command/login.rs
```diff
@@ -165,6 +165,7 @@ pub async fn login_command(_: XetConfig, args: &LoginArgs) -> errors::Result<()>
     }
     cfg.to_file(&global_config)
         .map_err(|e| errors::GitXetRepoError::ConfigError(e.into()))?;
+    println!("Login successful");
```
hoytak, line 168: Don't we want this to be eprintln ?

ylow: mmmm sure.

seanses (approved):

PR #159, opened 2023-10-20
Utility function to read and write directly to a bare repo

hoytak:
This PR adds simple utility functions to read and write files directly on a bare git repo.  

In this PR, these functions are not connected, but these are needed to enable some end-to-end tests that cover the different integration scenarios.

### DIFF
--- a/rust/gitxetcore/src/git_integration/git_wrap.rs
+++ b/rust/gitxetcore/src/git_integration/git_wrap.rs
@@ -1,12 +1,14 @@
 use crate::constants::MINIMUM_GIT_VERSION;
 use crate::errors::GitXetRepoError;
 use crate::errors::Result;
+use git2::Repository;
 use itertools::Itertools;
 use lazy_static::lazy_static;
 use regex::Regex;
 use std::path::PathBuf;
 use std::process::Command;
 use std::str::FromStr;
+use std::sync::Arc;
 use tracing::error;
 use tracing::{debug, info};
 use version_compare::Version;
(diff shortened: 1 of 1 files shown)


rust/gitxetcore/src/git_integration/git_wrap.rs
```diff
@@ -353,3 +355,181 @@ pub fn perform_git_version_check() -> Result<()> {
...
+    // Create a tree from the blobs; do this in memory, then do it one by one.
+    let mut index = git2::Index::new()?;
+
+    for (file_name, file_oid, data_len) in file_oids {
+        let entry = git2::IndexEntry {
+            path: file_name.as_bytes().into(),
+            id: file_oid,
+            file_size: data_len as u32,
+            mode: 0o100644, // represents a blob (file)
+            dev: 0,
```
hoytak, line 388: From https://github.com/libgit2/libgit2/blob/45fd9ed7ae1a9b74b957ef4f337bc3c8b3df01b5/tests/libgit2/reset/hard.c#L105, these are fine to leave as zero.

rust/gitxetcore/src/git_integration/git_wrap.rs
```diff
@@ -353,3 +355,181 @@ pub fn perform_git_version_check() -> Result<()> {
         }
     }
 }
+
+// Add files to a repo
```
seanses, line 359: Let's add a comment for the default behavior when branch_name is None.

rust/gitxetcore/src/git_integration/git_wrap.rs
```diff
@@ -353,3 +355,181 @@ pub fn perform_git_version_check() -> Result<()> {
...
+        &signature,
+        &signature,
+        commit_message,
+        &tree,
+        &parents,
+    )?;
+
+    Ok(())
+}
+pub fn read_file_from_repo(
```
seanses, line 439: Also here a comment for branch is None.

seanses (approved):
LG!

PR #161, opened 2023-10-20
Ensure git will not block on waiting for user input.

hoytak:
With integrating git with different endpoints, git will sometimes block on waiting for user input when the authentication configurations are not correct.  Since the integration methods probe several connection methods (https, ssh), this causes an issue.   This PR prevents this, and the git process will error out in such a case.

### DIFF
--- a/rust/gitxetcore/src/git_integration/git_wrap.rs
+++ b/rust/gitxetcore/src/git_integration/git_wrap.rs
@@ -66,6 +66,8 @@ fn spawn_git_command(
         &env
     );
 
+    // Using idea from https://stackoverflow.com/questions/30776520/closing-stdout-or-stdin
+
     // Disable stdin so it doesn't hang silently in the background.
     cmd.stdin(std::process::Stdio::piped());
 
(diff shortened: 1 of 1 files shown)


rust/gitxetcore/src/git_integration/git_wrap.rs
```diff
@@ -64,7 +66,27 @@ fn setup_git_command(
...
+    } else {
+        cmd.stdout(std::process::Stdio::inherit());
+        cmd.stderr(std::process::Stdio::inherit());
+    }
+
+    // Spawn the child
+    let mut child = cmd.spawn()?;
+
+    // Immediately drop the writing end of the stdin pipe; if git attempts to wait on stdin, it will cause an error.
+    drop(child.stdin.take());
```
seanses, line 87: Ha! This reminds me of building a shell in a course project

ylow:
Looks ok to me... Will let @seanses do the review.

seanses (approved):
Looks good if works!

PR #162, opened 2023-10-20
Upgrade checks will not fail if internet is down or API limit reached

hoytak:


seanses (approved):
LG! Merge main to get the test_dir_summary test fix in.

PR #166, opened 2023-10-20
Downgrade summary load error to info

znation:
After https://github.com/xetdata/xet-core/pull/154, the summary in git notes is no longer compatible with the struct layout for Serde serialization, which is by design (summaries will be regenerated). Since this is by design, this fix downgrades the error to an info message.

### DIFF
--- a/rust/gitxetcore/src/summaries_plumb.rs
+++ b/rust/gitxetcore/src/summaries_plumb.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use serde::{Deserialize, Serialize};
-use tracing::{debug, error, warn};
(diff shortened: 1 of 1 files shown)


seanses:
I thought after dropping the `file_type_simple_category` field from `LibmagicSummary` and parsing old summaries will just ignore that field, or is that only a JSON feature?

seanses (approved):

znation:
@seanses It is using Serde JSON serialization; not sure why dropping the field causes it to be unable to deserialize but it's fine to regenerate this summary when needed.

hoytak (approved):

PR #168, opened 2023-10-23
Adding printing path and logging the invocation command.

hoytak:
A simple update that: 
- If XET_PRINT_LOG_FILE_PATH, then the log path is dumped to stderr.
- After the log is set up, the invocation command of the program is logged as the first command.

### DIFF
--- a/rust/gitxetcore/src/config/util.rs
+++ b/rust/gitxetcore/src/config/util.rs
@@ -1,5 +1,6 @@
 use std::path::{Path, PathBuf};
 
+use itertools::Itertools;
 use xet_config::{Cas, Cfg, Log, User};
 
 use crate::command::CliOverrides;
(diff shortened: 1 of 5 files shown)


rust/gitxetcore/src/config/util.rs
```diff
@@ -126,6 +127,46 @@ impl<T: Sized> OptionHelpers<T> for Option<T> {
...
+
+    let cwd = std::env::current_dir()
+        .unwrap_or_default()
+        .to_str()
+        .unwrap_or_default()
+        .to_owned();
+
+    let subcommand: String = args
+        .map(|ee| {
+            let e = ee.strip_prefix(&cwd).unwrap_or(&ee);
```
seanses, line 158: May want to use Path::strip_prefix instead because str::strip_prefix leaves the forward slash in the result at the beginning.

seanses (approved):
LG! One nit comment above.

PR #169, opened 2023-10-23
Clear Initialization logic + More Init options

hoytak:
To enable the user integration scenarios, the init logic needed to be modified.  However, there was quite a bit of logic that had become very convoluted, making it difficult to properly trace what was going on.  In addition, to test these scenarios, we needed to add logic to the repository init fields to set up origin repos for these scenarios correctly.   

This PR organizes the init logic into two functions, perform_explicit_setup and perform_implicit_setup, and then moves the actions involved in these two into separate and clearly described linear tasks.  Overall, the number of various aggregate setup functions is reduced from 5-6 down to these two, making the logic much easier to follow.  

In addition, two options are added to the explicit setup function, namely a minimal flag, which only writes the basic commit files, and a an option to include the .xet/config.toml file as part of the initial commit.  These are needed for testing purposes. 
Furthermore, logic was added so that init can now work more fully on bare repos as well.


rust/gitxetcore/src/git_integration/git_repo.rs
```diff
@@ -618,101 +555,271 @@ impl GitRepo {
...
-            self.run_git_checked_in_repo(
-                "add",
-                &[self.repo_dir.join(".gitattributes").to_str().unwrap()],
-            )?;
-            self.run_git_checked("commit", &["-m", "Configured repository to use git-xet."])?;
+        if !is_bare && args.write_hooks {
+            repo_changed |= self.verify_or_write_hooks(args.enable_locking)?;
+        }
+
+        if args.write_remote_fetch_config {
```
seanses, line 584: maybe add `!is_bare` here

rust/gitxet/tests/integration_tests/test_git_xet_init.sh
```diff
@@ -108,10 +108,74 @@ mkdir repo_5
 pushd repo_5
 
 git init --bare
-git xet init -m 2 --force --bare
+git xet init -m 2 --force
```
seanses, line 111: Actually xetea runs `git xet init -m 2 --bare`, `--force` shouldn't be needed because it serves as a remote. Let's update this to reflect the same logic there.

rust/gitxet/tests/integration_tests/test_git_xet_init.sh
```diff
@@ -108,10 +108,74 @@ mkdir repo_5
...
+
+
+echo "[config]" >> config.toml
+echo "url = \"$(pwd)/repo_9\"" >> config.toml
+
+
+mkdir repo_9
+pushd repo_9
+git init --bare 
+git xet init -m 2 --force --explicit --write-gitattributes --write-repo-salt --xet-config-file=../config.toml
```
seanses, line 162: Add drops `--force` here.

rust/gitxet/tests/integration_tests/test_git_xet_init.sh
```diff
@@ -108,10 +108,74 @@ mkdir repo_5
...
+[[ -e .git/refs/notes/xet/reposalt ]] || die "reposalt not set"
+[[ -e .git/refs/notes/xet/merkledb ]] || die "merkledb v1 guard notes not set"
+[[ -e .git/refs/notes/xet/merkledbv2 ]] || die "merkledb v2 notes not set"
+[[ -e .gitattributes ]] || die ".gitattributes not created"
+popd
+
+mkdir repo_7
+pushd repo_7
+git init --bare 
+git xet init -m 2 --force --explicit --write-gitattributes --write-repo-salt
```
seanses, line 137: Similarly drops `--force` here.

seanses (approved):
Some comments above to drop unneeded flags in tests. The rest looks good!

rust/gitxet/tests/integration_tests/test_git_xet_init.sh
hoytak: Ah, good catch.  I'll update force so that if it runs on a bare repo, it doesn't do the check for remotes.

PR #170, opened 2023-10-24
Updated logic and error message printing in probing github remote.

hoytak:
Updated how we probe github remotes: 
- If the remote URL formed from the .xet/config.toml file is one of the remotes, then just probe the remote, not both. 
- If one version of https or ssh succeeds, then don't probe the other one. 
- If there is an error, don't log an error -- just swallow it.


seanses (approved):
LG!

PR #173, opened 2023-10-24
fix write lazyconfig on clone

seanses:
Moving writing the lazyconfig from clone command to filter command because we cannot reliably detect the cloned directory.
Also adds the lazy logic into v1 repo path.
Fix https://github.com/xetdata/xethub/issues/4427

### DIFF
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -1,9 +1,7 @@
-use std::env::current_dir;
-
 use crate::constants::GIT_LAZY_CHECKOUT_CONFIG;
 use crate::git_integration::git_repo::{verify_user_config, GitRepo};
 use clap::Args;
-use lazy::lazy_config::write_default_lazy_config;
+use lazy::lazy_config::XET_LAZY_CLONE_ENV;
 
 use crate::config::XetConfig;
 use crate::errors::Result;
@@ -34,7 +32,7 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
     verify_user_config(None)?;
     eprintln!("Preparing to clone Xet repository.");
 
-    let (repo, _) = GitRepo::clone(
+    GitRepo::clone(
         Some(&config),
         &arg_v[..],
         args.no_smudge || args.lazy,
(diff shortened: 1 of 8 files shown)


rust/gitxetcore/src/data_processing_v2.rs
```diff
@@ -956,7 +956,13 @@ impl PointerFileTranslatorV2 {
         match fi {
             Some(ptr) => {
                 if let Some(lazy) = &self.lazyconfig {
-                    let rule = lazy.match_rule(path).unwrap();
+                    let rule = lazy
+                        .match_rule(path)
+                        .map_err(|e| {
+                            eprintln!("LazyConfig error matching rule: {e:?}");
```
hoytak, line 962: Let's dump this to error! instead.

rust/gitxetcore/src/data_processing_v1.rs
```diff
@@ -639,6 +650,23 @@ impl PointerFileTranslatorV1 {
 
         match fi {
             Some(ptr) => {
+                if let Some(lazy) = &self.lazyconfig {
+                    let rule = lazy
+                        .match_rule(path)
+                        .map_err(|e| {
+                            eprintln!("LazyConfig error matching rule: {e:?}");
```
hoytak, line 657: Let's dump this to error instead.

rust/gitxetcore/src/command/clone.rs
```diff
@@ -34,7 +32,24 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
...
+    if args.lazy {
+        // We cannot reliably detect the directory cloned into,
+        // because of how git picks the "humanish" part of the source
+        // repository (https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-ltdirectorygt),
+        // and scenarios when a directory is specified in command line arguments.
+        //
+        // So we set a env var and delegate setting the lazy configuration
+        // to the filter command.
+
+        eprintln!(
```
hoytak, line 44: Let's also log this (thinking of the scenario down the road where we are only able to use user's logs to debug things).

rust/gitxetcore/src/data_processing_v1.rs
```diff
@@ -639,6 +650,23 @@ impl PointerFileTranslatorV1 {
...
+                            e
+                        })
+                        .unwrap();
```
hoytak, line 660: I'm a bit worried about the potential panic here -- would it make sense to do unwrap_or(...) and have a default?

rust/lazy/src/lazy_config.rs
```diff
@@ -93,9 +97,17 @@ impl LazyConfig {
...
+/// Check if the config file exists and if parsing succeeds.
+/// Otherwise write the default config.
+pub async fn check_or_write_default_lazy_config(config_file: &Path) -> Result<()> {
+    if config_file.is_file() {
+        LazyConfig::load_from_file(config_file).await?;
+        return Ok(());
+    }
 
-pub fn write_default_lazy_config(config_file: &Path) -> Result<()> {
+    debug!("Writing \"{DEFAULT_LAZY_RULE}\" to {config_file:?}");
```
hoytak, line 110: Could be info! level?

rust/gitxetcore/src/data_processing_v1.rs
seanses: Let default to POINTER because a lazy config exists.

rust/gitxetcore/src/data_processing_v1.rs
seanses: The error `e` already contains the path so I think it's better to not repeat it.

seanses:
Superseded by https://github.com/xetdata/xet-core/pull/183.

PR #177, opened 2023-11-01
Fix for git xet init issue on bare repos

hoytak:
The recently added git xet init functionality is actually broken; this PR fixes it.  

The create_commit function actually discards all previous commits.  This is due to creating a blank index instead of a tree as an intermediate step, then writing this as the commit; however, this sets the current state of the repo, not an update. 

The solution is to use a new, more general atomic commit function to make this happen.   A number of tests are added to cover this scenario.


seanses (approved):
LG!

PR #178, opened 2023-11-02
Fix summary deserialization compat

znation:
The PR #154 broke summary deserialization by removing a field from the LibmagicSummary type. This was intentional for libmagic summaries, but I didn't realize this would also affect deserialization of CSV summaries since we combine them together in a WholeRepoSummary struct that we then try to deserialize. We still want to be able to read CSV summaries through that struct, so we need to keep this field forever for Serde compat.


hoytak (approved):
Looks good to me.

jgodlew (approved):
:shipit:

PR #180, opened 2023-11-02
Update to git xet clone to allow stdin.

hoytak:


rust/gitxetcore/src/command/mount.rs
```diff
@@ -282,6 +282,7 @@ If you use a git UI, point it to the raw path.
             false,             // no smudge
             Some(&clone_path), // base dir
             false,             // passthrough
+            false,
```
seanses, line 285: Maybe allow stdin here because we can mount github repos.

hoytak: So looking at this -- we don't actually pass through stdout or stderr on those ; everything is captured.  So I don't think that allowing stdin there makes sense.

seanses (approved):

PR #181, opened 2023-11-02
Disable version check in recursive calls.

hoytak:
Disables the version check in all calls spawned downstream from another git-xet process.   Addresses https://github.com/xetdata/xethub/issues/4600 and other cases where the version check message may be swallowed silently due to this.


seanses (approved):
LG!

PR #182, opened 2023-11-06
List files from git repo

seanses:
This implements a helper function to list files under a path in a repo, needed by git xet materialize and dematerialize features.

PR #183, opened 2023-11-06
Simplify lazy

seanses:
Simplify the lazy logic to use a list of file paths as rules. Introduce interactice commands `materialize` and `dematerialize`. Based on https://www.notion.so/xethub/Lazy-clone-docs-8afba2483efd4696b9f0c154754b3536?pvs=4#23617a64ffb8459c8cbe7c1e74fb0d7c. Depends on #182.

### DIFF
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -1,7 +1,7 @@
 use crate::constants::GIT_LAZY_CHECKOUT_CONFIG;
 use crate::git_integration::git_repo::{verify_user_config, GitRepo};
 use clap::Args;
-use lazy::lazy_config::XET_LAZY_CLONE_ENV;
+use lazy::lazy_rule_config::XET_LAZY_CLONE_ENV;
 use tracing::info;
 
 use crate::config::XetConfig;
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -1,8 +1,5 @@
-use crate::constants::GIT_LAZY_CHECKOUT_CONFIG;
 use crate::git_integration::git_repo::{verify_user_config, GitRepo};
 use clap::Args;
-use lazy::XET_LAZY_CLONE_ENV;
-use tracing::info;
 
 use crate::config::XetConfig;
 use crate::errors::Result;
@@ -32,26 +29,6 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
     // First check local config.
     verify_user_config(None)?;
     eprintln!("Preparing to clone Xet repository.");
-
-    if args.lazy {
-        // We cannot reliably detect the directory cloned into,
-        // because of how git picks the "humanish" part of the source
-        // repository (https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-ltdirectorygt),
-        // and scenarios when a directory is specified in command line arguments.
-        //
-        // So we set a env var and delegate setting the lazy configuration
-        // to the filter command.
-
-        eprintln!(
-            "Setting up lazyconfig under your repo at REPO_ROOT/{GIT_LAZY_CHECKOUT_CONFIG}. 
-If cloning an empty branch, please create this file manually at the aforementioned path."
-        );
-        info!(
-            "Setting up lazyconfig under the cloned repo at REPO_ROOT/{GIT_LAZY_CHECKOUT_CONFIG}"
-        );
-
-        std::env::set_var(XET_LAZY_CLONE_ENV, "1");
-    }
 
     GitRepo::clone(
         Some(&config),
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -1,9 +1,7 @@
-use std::env::current_dir;
-
 use crate::constants::GIT_LAZY_CHECKOUT_CONFIG;
 use crate::git_integration::git_repo::{verify_user_config, GitRepo};
 use clap::Args;
-use lazy::lazy_config::write_default_lazy_config;
+use lazy::lazy_config::XET_LAZY_CLONE_ENV;
 
 use crate::config::XetConfig;
 use crate::errors::Result;
@@ -34,7 +32,7 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
     verify_user_config(None)?;
     eprintln!("Preparing to clone Xet repository.");
 
-    let (repo, _) = GitRepo::clone(
+    GitRepo::clone(
         Some(&config),
         &arg_v[..],
         args.no_smudge || args.lazy,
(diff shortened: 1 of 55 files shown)


rust/gitxetcore/src/command/lazy.rs
```diff
@@ -39,6 +41,7 @@ impl LazyCommandShim {
             LazyCommand::Check => "check".to_string(),
             LazyCommand::Match(_) => "match".to_string(),
             LazyCommand::Apply => "apply".to_string(),
+            LazyCommand::Print => "print".to_string(),
```
hoytak, line 44: Instead of Print/print, let's use Show/show -- (like the git show command)

rust/gitxetcore/src/command/materialize.rs
```diff
@@ -0,0 +1,60 @@
...
+    recursive: bool,
+
+    path: String,
+}
+
+pub async fn materialize_command(cfg: XetConfig, args: &MaterializeArgs) -> Result<()> {
+    // Make sure repo working directory is clean
+    let repo = GitRepo::open(cfg.clone())?;
+
+    eprintln!("Checking repo state...");
```
hoytak, line 21: This command should be nearly instantaneous, so I think we can drop the print statement here (we don't print anything elsewhere when we check).

seanses: Sounds good will remove.

rust/gitxetcore/src/command/materialize.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let lazy_config_path = if let Some(path) = &cfg.lazy_config {
+        path.to_owned()
+    } else {
+        let path = cfg.repo_path()?.join(GIT_LAZY_CHECKOUT_CONFIG);
+        check_or_create_lazy_config(&path).await?;
+        path
+    };
+
+    eprintln!("Materializing...");
```
hoytak, line 37: Let's move this later, like right after the path_list parsing, and add some information about the input (e.g. "Materializing {} files.").

rust/gitxetcore/src/command/materialize.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let path_list_ref: Vec<_> = path_list.iter().map(|s| s.as_str()).collect();
+
+    lazyconfig.add_rules(&path_list_ref);
+
+    // saves changes to the file
+    drop(lazyconfig);
+
+    // rerun smudge filter
+    std::fs::remove_file(cfg.repo_path()?.join("index"))?;
```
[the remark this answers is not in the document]
seanses: Unfortunately yes.. I intended to use `checkout -- [pathspec]` but didn't find a way to force re-filter on a subset of files.

rust/lazy/src/lazy_pathlist_config.rs
```diff
@@ -0,0 +1,285 @@
...
+            }
+            // comments
+            if line_content.starts_with('#') {
+                continue;
+            }
+
+            if !pathlist.insert(line_content.to_owned()) {
+                // already contained this line
+                if warns_duplicate {
+                    println!("Found duplicate path {line_content:?} in the list.");
```
hoytak, line 45: Let's change this to warn! (println! will mess up parsed outputs)

rust/lazy/src/lazy_pathlist_config.rs
```diff
@@ -0,0 +1,285 @@
...
+    ops::{Deref, DerefMut},
+    path::{Path, PathBuf},
+};
+
+use tracing::{error, info};
+
+use crate::{error::Result, lazy_rule::LazyStrategy};
+
+#[derive(Debug)]
+pub struct LazyPathListConfig {
```
hoytak, line 14: Small thing, but could save time on larger repos -- we could do an optimization here by saving one large String that is the entire loaded file, then we store a HashSet of &str objects with the same lifetime (and could also then count newlines and use HashSet::with_capacity).

ylow: How about file names with newlines?

rust/gitxetcore/src/command/materialize.rs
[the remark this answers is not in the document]
ylow: Do we even need to rebuild the index?

rust/lazy/src/lazy_pathlist_config.rs
ylow: This seems like a poor micro-optimization with no clear benefits.

rust/gitxetcore/src/command/clone.rs
```diff
@@ -34,7 +33,27 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
...
+        // We cannot reliably detect the directory cloned into,
+        // because of how git picks the "humanish" part of the source
+        // repository (https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-ltdirectorygt),
+        // and scenarios when a directory is specified in command line arguments.
+        //
+        // So we set a env var and delegate setting the lazy configuration
+        // to the filter command.
+
+        eprintln!(
+            "Setting up lazyconfig under your repo at REPO_ROOT/{GIT_LAZY_CHECKOUT_CONFIG}.
```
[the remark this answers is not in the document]
seanses: Good point. This was a side effect from the previous implementation. Should not need it now.

rust/lazy/src/lazy_pathlist_config.rs
hoytak, line 14: Deleted original comment, but acknowledge.

rust/gitxetcore/src/command/materialize.rs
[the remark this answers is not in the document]
hoytak: If the difference is in the filter behavior, git won't update the working tree otherwise.  However, there are likely ways around this.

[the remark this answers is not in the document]
hoytak: Try this (using git2)

```
   // Get the index
    let mut index = repo.index()?;

    // Add the file to the index, forcing a re-checkout
    // This step doesn't strictly re-checkout the file, but it's a necessary precursor
    // to update the index and ensure that the checkout process will actually move the file.
    index.add_path(Path::new(file_path))?;
    index.write()?;

    // Perform a checkout of the file from the index to the working directory
    // This should invoke any filter processes such as LFS or others defined.
    repo.checkout_index(Some(&mut index), Some(git2::build::CheckoutBuilder::new().force()))?;
```

hoytak (approved):
LGTM after small things.

rust/gitxetcore/src/command/materialize.rs
seanses: Also added message for no file found under the input path.

rust/gitxetcore/src/command/clone.rs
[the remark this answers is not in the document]
seanses: Dropped.

rust/gitxetcore/src/command/materialize.rs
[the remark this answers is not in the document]
seanses: Tried, doesn't work :(
```
// Get the index
    let mut index = repo.repo.index()?;

    // Add the file to the index, forcing a re-checkout
    // This step doesn't strictly re-checkout the file, but it's a necessary precursor
    // to update the index and ensure that the checkout process will actually move the file.
    for p in path_list_ref {
        index.add_path(std::path::Path::new(p))?;
    }
    index.write()?;

    // Perform a checkout of the file from the index to the working directory
    // This should invoke any filter processes such as LFS or others defined.
    repo.repo.checkout_index(
        Some(&mut index),
        Some(git2::build::CheckoutBuilder::new().force()),
    )?;
```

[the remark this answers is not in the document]
ylow: what happens if we don't rebuild the index?

rust/gitxetcore/src/command/materialize.rs
```diff
@@ -56,11 +77,37 @@ pub async fn materialize_command(cfg: XetConfig, args: &MaterializeArgs) -> Resu
...
+                        "Failed to get relative path for {p:?} to the current working directory."
+                    );
+                    GitXetRepoError::Other(e.to_string())
+                })?;
+        touch_file(relative_path_from_curdir)?;
+        git_args.push(relative_path_from_curdir.to_str().unwrap_or_default());
+    }
 
-    git_wrap::run_git_captured(None, "checkout", &["--force"], true, None)?;
+    git_wrap::run_git_captured(None, "checkout", &git_args, true, None)?;
```
ylow, line 97: ok, but I worry git_args can get long.

ylow: There is pathspec-from-file which can read from a file or stdin, but it seems to exist only as of 2.27.

PR #187, opened 2023-11-07
Push repo salt in case the remote is not xet-enabled

seanses:
In our tests the remote is not xet-enabled and a local clone set up the salt.


hoytak (approved):
Per discussion, decided this should actually be in here.

PR #189, opened 2023-11-07
Fix for intermediate test failure.

hoytak:
git commit seems to barf out an exit code 1 sometimes if it thinks that the file hasn't changed from the index based on timestamps.  This can sometimes causes the test_github_integration_simple test to intermittently fail depending on where the boundary of the 1 second tick used for difference checking.   This PR fixes this.


seanses (approved):
LG! Annoying git miscellanies...

PR #190, opened 2023-11-07
Pad out spaces on progress monitoring. 

hoytak:
Puts a minimum width for the line we print in our progress bar, with the difference made up with spaces.  This is based on git's messages, which can be longer than ours, causing weird interleaving when the user does stuff to the input (hits a new line, etc.)  .

### DIFF
--- a/rust/progress_reporting/src/data_progress.rs
+++ b/rust/progress_reporting/src/data_progress.rs
@@ -143,6 +143,11 @@ impl DataProgressReporter {
 
     /// Does the actual printing
     fn print(&self, is_final: bool) -> std::result::Result<(), std::io::Error> {
+        // Put a minimum width for the line we print.  This is based on git's messages,
+        // which can be longer than ours, causing weird inteleaving when the user does
+        // stuff to the input (hits a new line, etc.)  So, print a minimum buffer of spaces.
+        const PRINT_LINE_MIN_WIDTH: usize = 74;
+
         if self.disable || !self.is_active.load(Ordering::Relaxed) {
             return Ok(());
         }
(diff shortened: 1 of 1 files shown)


rust/progress_reporting/src/data_progress.rs
```diff
@@ -297,8 +304,10 @@ impl DataProgressReporter {
             }
         };
 
-        if write_str.len() < lg_print_info.last_write_length {
-            let mut len_to_write = lg_print_info.last_write_length - write_str.len();
+        let write_str_pad_len = lg_print_info.last_write_length.min(PRINT_LINE_MIN_WIDTH);
```
seanses, line 307: .max?

rust/progress_reporting/src/data_progress.rs
```diff
@@ -143,6 +143,11 @@ impl DataProgressReporter {
 
     /// Does the actual printing
     fn print(&self, is_final: bool) -> std::result::Result<(), std::io::Error> {
+        // Put a minimum width for the line we print.  This is based on git's messages,
+        // which can be longer than ours, causing weird inteleaving when the user does
+        // stuff to the input (hits a new line, etc.)  So, print a minimum buffer of spaces.
+        const PRINT_LINE_MIN_WIDTH: usize = 74;
```
seanses, line 149: Wondering why choosing 74

rust/progress_reporting/src/data_progress.rs
hoytak: Oh bleh.  Great catch.

rust/progress_reporting/src/data_progress.rs
hoytak: Measured the length of the git progress strings that the original bug report showed.

seanses (approved):
LGTM!

PR #192, opened 2023-11-08
Refactored git_integration.

hoytak:
The git_integration/ directory was getting problematic; everything was being jammed in to git_repo.rs or git_wrap.rs, without much logic about what was in each of those files.  This PR simply moves the functionality in those two files into logically defined source files and cleans up the imports in this directory.  It touches a lot, but there should be no functionality change.


rust/gitxetcore/src/git_integration/command_entry.rs
seanses, line 1: Can we name this `hook_command_entry`? From my past experience I always tried to find this file using "hook" as the file name search keyword .

rust/gitxetcore/src/git_integration/mod.rs
```diff
@@ -1,7 +1,22 @@
...
+mod git_repo_plumbing;
+pub mod git_repo_salt;
+mod git_xet_repo;
+
 pub mod git_url;
-pub mod git_wrap;
+pub mod git_user_config;
+pub mod git_version_checks;
+
+pub use crate::git_integration::git_xet_repo::git_repo_test_tools; // HERE
```
seanses, line 16: Can we drop the `crate::git_integration::` in the beginning. Also, I'm thinking if we can drop all `pub` from mod [...] above and re-export their internal functions so imports in other source files can be concise.

seanses (approved):
Thanks for refactoring! Overall looks good to me. Some suggestions above, mainly a personal habit thing.

rust/gitxetcore/src/git_integration/mod.rs
hoytak: I tried doing just that, but there are a bunch of places (like the version checks) that are only called in one place, and the names beneath them aren't very precise.  So I went back to doing it like this for the lesser used modules.  We can still use git_integration::*; then e.g. git_user_config::verify_users to call them.

PR #193, opened 2023-11-10
Swtich to embedded TLS roots instead of Native TLS

ylow:
Mac Native TLS takes a long time to load.
https://github.com/rustls/rustls-native-certs/issues/30

This is an issue as some git operations spawn (sequentially) a large number of git-xet subprocesses. This switches reqwest to using an embedded Mozilla certificate roots.  (webpki-roots crate). This reduces push time by about 33%.

### DIFF
--- a/rust/gitxetcore/Cargo.toml
+++ b/rust/gitxetcore/Cargo.toml
@@ -96,7 +96,7 @@ mockall_double = "0.3"
 
 # axe
 sysinfo = "0.26.6"
-reqwest = {version = "0.11.4", features = ["json"]}
+reqwest = {version = "0.11.4", features = ["json", "webpki-roots"]}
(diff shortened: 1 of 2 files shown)


znation (approved):
Good find!

alerque:
Could this be made optional? Embedding root certificates (even for internal use) is a big no-no for some Linux distributions. I package this for Arch Linux for example which has a policy against it. Since the problem is macOS specific could the option to eschew the system certificates at least be made platform specific? If this isn't a problem on Linux but the "solution" is a problem in its own right, this is a step backwards.

ylow:
#195 amends this to use embedded root certs only on MacOS.

alerque:
Thanks for that!

PR #194, opened 2023-11-10
Disable recursive reference transaction hooks

hoytak:
This PR disables recursive reference transaction hooks in the MerkleDB v2 case.  It is unclear why they are needed in the MDB v1 case; this warrants further investigation.

On the pre-push hook, this seems to save roughly 1.5 - 2 seconds of time depending on the scenario.


ylow:
Do we have the correct mdb version at the point when the hooks are written? Thinking about other initialization conditions where we may have an incomplete information about the repository.

hoytak:
Yes. In both the implicit and explicit paths, the hooks are set up last.

ylow (approved):

seanses:
How much time does this save approximately? Should be worthy documenting that somewhere, either in the code or in the PR comment.

seanses (approved):

PR #196, opened 2023-11-10
Plumbing functions for walking working directory

seanses:
Implements two functions: `walk_working_dir` for walking a repo working directory; `filter_files_from_index` for filtering out untracked files. Needed for `materialize` and `dematerialize` commands.

### DIFF
--- a/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
+++ b/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
@@ -834,6 +834,7 @@ mod git_repo_tests {
 
     #[test]
     #[serial(set_curdir)] // tests messing up with 'set_current_dir' run serial
+    #[ignore] // however, tests with neither #[serial] nor #[parallel] may run at any time, thus to ignore in CI to avoid occasional errors.
     fn test_walk_working_dir() -> anyhow::Result<()> {
         let repo = TestRepo::new()?;
 
@@ -956,6 +957,7 @@ mod git_repo_tests {
 
     #[test]
     #[serial(set_curdir)] // tests messing up with 'set_current_dir' run serial
+    #[ignore] // however, tests with neither #[serial] nor #[parallel] may run at any time, thus to ignore in CI to avoid occasional errors.
     fn test_filtered_untracked_files() -> anyhow::Result<()> {
         let repo = TestRepo::new()?;
 
--- a/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
+++ b/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
@@ -454,7 +454,7 @@ pub fn filter_files_from_index(
 }
 
 pub fn is_git_special_files(path: &str) -> bool {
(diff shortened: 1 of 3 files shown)


ylow:
If helpful we have used the walkdir package elsewhere https://docs.rs/walkdir/latest/walkdir/

seanses:
> If helpful we have used the walkdir package elsewhere https://docs.rs/walkdir/latest/walkdir/

Sure. Simple change.

rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
```diff
@@ -357,14 +359,148 @@ pub fn list_files_from_repo(
...
+
+        let dir_walker = std::fs::read_dir(sr)?;
+
+        for entry in dir_walker.flatten() {
+            let file_type = entry.file_type()?;
+            let file_path = entry.path();
+
+            if file_type.is_file() {
+                ret.push(file_path.strip_prefix(&work_root)?.to_owned());
+            } else if recursive
```
hoytak: I think we need a check here for is_dir(); if it's a symlink, it ends up getting put on the queue and treated as a directory.

seanses: Good point. I've rewritten this part using walkdir, should exclude symlinks.

rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
```diff
@@ -357,14 +359,135 @@ pub fn list_files_from_repo(
...
+        return Ok(vec![search_root.strip_prefix(work_root)?.to_owned()]);
+    }
+
+    // now search_root is a regular directory, walk below it
+
+    let ret: Vec<_> = WalkDir::new(&search_root)
+        .follow_links(false)
+        .max_depth(if recursive { usize::MAX } else { 1 })
+        .into_iter()
+        .filter_entry(|e| !is_git_special_files(e.file_name().to_str().unwrap_or_default()))
```
hoytak, line 413: Does this actually exclude the files under .git/ with the walkdir iter now?  It has to be an exact match, so I think the is_git_special_files function needs to test the prefix.

hoytak: Ah, I see that it does.  Nice.

rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
```diff
@@ -357,14 +359,135 @@ pub fn list_files_from_repo(
...
+            assert_eq!(bytes2path(&_entry.path), f);
+            ret.push(f.to_owned());
+        }
+    }
+
+    Ok(ret)
+}
+
+pub fn is_git_special_files(path: &str) -> bool {
+    matches!(path, ".git" | ".gitignore" | ".gitattributes")
```
hoytak, line 457: Do we also want to add .xet here?

hoytak (approved):
Looks good to me!  Minor things.

PR #199, opened 2023-11-14
Speedup of pre-push hook

hoytak:
This PR speeds up the pre-push hook significantly, cutting down the pre-push hook execution time on the Flickr30k dataset, with one change committed, from 9.5 seconds to 5 seconds.   There are two main parts:

1. Initializing the CAS once, and using it for all operations.   The previous version create a CAS staging client three independent times.  Each time takes well over a second due to the time required to scan the cache folder (another issue, the next one to look at).

2. Running the operations in parallel while respecting the dependencies.  The uploading to the CAS is done in the background for all of it.  Uploading the shards is done in the background until the notes are needed.

### DIFF
--- a/rust/gitxetcore/src/errors.rs
+++ b/rust/gitxetcore/src/errors.rs
@@ -106,6 +106,9 @@ pub enum GitXetRepoError {
     #[error("Repo Salt Unavailable: {0}")]
     RepoSaltUnavailable(String),
 
+    #[error("Subtask scheduling error: {0}")]
+    JoinError(#[from] tokio::task::JoinError),
+
     #[error("Lazy Config Error : {0}")]
     LazyConfigError(#[from] LazyError),
 }
@@ -162,6 +165,7 @@ impl From<GitXetRepoError> for ExitCode {
             GitXetRepoError::RepoUninitialized(_) => 30,
             GitXetRepoError::RepoSaltUnavailable(_) => 31,
             GitXetRepoError::LazyConfigError(_) => 30,
+            GitXetRepoError::JoinError(_) => 32,
         })
     }
 }
--- a/rust/gitxetcore/src/git_integration/git_xet_repo.rs
+++ b/rust/gitxetcore/src/git_integration/git_xet_repo.rs
@@ -1705,6 +1705,19 @@ impl GitXetRepo {
                     GIT_NOTES_MERKLEDB_V2_REF_NAME,
                 )?;
 
+                // Make sure that all the uploads and everything are in a good state before proceeding with
+                // anything changing the remote repository.
+                //
+                // Waiting until the CAS uploads finish avoids the following scenario:
+                // 1. user 1 commit file A and push, but network drops after
+                // sync_notes_to_remote before uploading cas finishes.
+                // 2. user 2 tries to git add the same file A, which on filter pulls in
+                // the new notes, and file A is 100% deduped so no CAS blocks will be created,
+                // and push.
+                //
+                // This results in a bad repo state.
+                upload_all_jh.await??;
+
                 self.sync_notes_to_remote(remote)?;
 
                 // Finally, we can move all the mdb shards from the session directory, which is used
@@ -1714,8 +1727,6 @@ impl GitXetRepo {
                     &self.merkledb_v2_cache_dir,
                 )
                 .await?;
-
-                upload_all_jh.await??;
             }
         };
         Ok(())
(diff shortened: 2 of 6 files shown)


rust/gitxetcore/src/errors.rs
```diff
@@ -162,6 +165,7 @@ impl From<GitXetRepoError> for ExitCode {
             GitXetRepoError::RepoUninitialized(_) => 30,
             GitXetRepoError::RepoSaltUnavailable(_) => 31,
             GitXetRepoError::LazyConfigError(_) => 30,
```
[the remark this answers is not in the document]
hoytak: Hmmm.  Good point.  Separate PR?

rust/gitxetcore/src/merkledb_shard_plumb.rs
```diff
@@ -606,7 +619,9 @@ pub async fn sync_session_shards_to_remote(
             git_xet_version: crate::data_processing_v2::GIT_XET_VERION.to_string(),
         };
 
-        let shard_file_client = shard_client::from_config(shard_connection_config).await?;
+        let shard_file_client = shard_client::from_config(shard_connection_config)
+            .await
+            .expect("1");
```
seanses, line 624: Why `expect("1")` here?

rust/gitxetcore/src/merkledb_shard_plumb.rs
```diff
@@ -620,17 +635,16 @@ pub async fn sync_session_shards_to_remote(
                 "Uploading shard {shard_prefix_ref}/{:?} from staging area to CAS.",
                 &si.shard_hash
             );
-            let data = fs::read(&si.path)?;
+            let data = fs::read(&si.path).expect("2");
```
seanses, line 638: And `expect("2")` here?

rust/gitxetcore/src/merkledb_shard_plumb.rs
hoytak: Ah, was debugging.  good catch.

rust/gitxetcore/src/merkledb_shard_plumb.rs
```diff
@@ -549,16 +550,12 @@ fn convert_merklememdb(
 
 /// Consolidate shards from sessions, install guard into ref notes v1 if
 /// a conversion was done. Write shards into ref notes v2.
-pub async fn sync_mdb_shards_to_git(
+pub async fn sync_session_mdb_shards_to_git(
```
seanses, line 553: I prefer not to wrap `update_mdb_shards_to_git_notes` and `move_session_shards_to_local_cache` into a single function. `sync_session_mdb_shards_to_git` is only used once in this file and the name doesn't describe all the operations inside.

rust/gitxetcore/src/git_integration/git_xet_repo.rs
```diff
@@ -1575,56 +1587,137 @@ impl GitXetRepo {
...
     }
 
     /// Pushes all the staged data in the local CAS
     pub async fn upload_all_staged(&self) -> Result<()> {
-        let repo = PointerFileTranslator::from_config(&self.xet_config).await?;
-        repo.upload_cas_staged(false).await?;
-        Ok(())
+        let cas = self.get_staging_cas().await?;
+
+        cas.upload_all_staged(MAX_CONCURRENT_UPLOADS, true)
```
seanses, line 1625: We don't want to retain uploaded staged cas blocks.

rust/gitxetcore/src/merkledb_shard_plumb.rs
hoytak: Agreed.  This was an intermediate function involved in how I was trying to split out the functionality into background and foreground stuff in the prepush hook.  Will fix.

rust/gitxetcore/src/git_integration/git_xet_repo.rs
hoytak: Ah.  Good catch.  Changed.

rust/gitxetcore/src/git_integration/git_xet_repo.rs
```diff
@@ -1575,56 +1587,137 @@ impl GitXetRepo {
...
+
+                // Finally, we can move all the mdb shards from the session directory, which is used
+                // by the upload_shard task, to the cache.
+                move_session_shards_to_local_cache(
+                    &self.merkledb_v2_session_dir,
+                    &self.merkledb_v2_cache_dir,
+                )
+                .await?;
 
+                upload_all_jh.await??;
```
seanses, line 1718: Move this to before `sync_notes_to_remote` so there's no dependency issue in the remote?

hoytak: In this case, the upload all task is just uploading all of the staged CAS blocks from the files. So it is independent of any of the notes.

seanses (approved):
LGTM!

PR #203, opened 2023-11-16
Avoided loading file in cache to extract block id.

hoytak:
In our cache, the key is encoded in the filename and comprised of the CAS name, the block index, and the block size.  However, on loading the cache, this information is read from the file instead of extracted from the file name.   On my local cache of roughly 800 files, this cuts the init time for the CAS cache from 1.5 seconds to ~100 ms.


rust/cache/src/disk/storage.rs
```diff
@@ -230,16 +230,21 @@ fn to_cache_value(entry: DirEntry) -> Result<CacheValue, String> {
...
 
     let file_size = metadata_size(&metadata) - header.get_header_len();
 
     Ok(CacheValue {
         size: file_size,
         version: 1,
-        key,
         block_size: header.block_size,
         block_idx: header.block_idx,
+        key: header.key,
```
seanses, line 247: If `Header::attempt_from_key(key)` somehow fails and header is set from `verify_header`, then `header.key` is only the `request.metadata().name()` instead of the entire `"{}.{}.{}",
        request.metadata().name(),
        request.block_range().idx(),
        request.block_size(),`

hoytak: From what I can trace, the name is the format!(...) thing.  This took a while to realize.

seanses (approved):
LGTM!

PR #205, opened 2023-11-18
Robust .xet permission handling

seanses:
Adds a `xet_home` in XetConfig. Each xet metadata folder/file to access (cache, repos, version_check_info), if not explicitly configured to a different path, will be created or read/written under a checked `xet_home`. Such directories and files will also be checked or created for R+W permissions with `Permission::create_dir_all` or `Permission::create_file`. If the running git-xet process lacks such permission, an error message is printed asking the user to reset the permissions.

Fix https://github.com/xetdata/xethub/issues/4592

### DIFF
--- a/rust/gitxetcore/src/config/errors.rs
+++ b/rust/gitxetcore/src/config/errors.rs
@@ -109,6 +109,6 @@ pub enum ConfigError {
     #[error("Error parsing local repo config file: {0}")]
     RepoConfigFileParseError(String),
 
-    #[error("Lack of read-write permission for path: {0}")]
-    LackofPermission(PathBuf),
+    #[error("Cound not find parent for path: {0}")]
+    InvalidPathParent(PathBuf),
 }
--- a/rust/gitxetcore/src/config/errors.rs
+++ b/rust/gitxetcore/src/config/errors.rs
@@ -108,4 +108,7 @@ pub enum ConfigError {
 
     #[error("Error parsing local repo config file: {0}")]
     RepoConfigFileParseError(String),
+
+    #[error("Lack of read-write permission for path: {0}")]
+    LackofPermission(PathBuf),
 }
--- a/rust/gitxetcore/src/config/permission.rs
+++ b/rust/gitxetcore/src/config/permission.rs
@@ -1,10 +1,8 @@
+use anyhow::anyhow;
 use colored::Colorize;
-#[cfg(unix)]
-use libc;
-use std::path::Path;
-#[cfg(windows)]
-use std::ptr;
-use tracing::warn;
+use rand::{thread_rng, Rng};
+use std::path::{Path, PathBuf};
+use tracing::info;
 #[cfg(windows)]
 use winapi::{
     shared::winerror::ERROR_SUCCESS,
@@ -16,10 +14,29 @@ use winapi::{
     },
 };
 
-#[derive(Debug, Clone)]
+// Facts:
+// Assume there's a standard user A that is not a root user.
+// 1. On Unix systems, suppose there is a path 'dir/f' where 'dir' is created by A but 'f'
+//    created by 'sudo A', A can read, rename or remove 'dir/f'. This implies that it's enough
+//    to check the permission of 'dir' if we don't directly write into 'dir/f'. This is exactly
+//    how we interact with the xorb cache: if an eviction is deemed necessary, the replacement
+//    data is written to a tempfile first and then renamed to the to-be-evicted entry. So even
+//    if a certain cache file was created by 'sudo A', the eviction by 'A' will succeed.
+// 2. On Windows, 'Run as administrator' by logged in user A actually sets %HOMEPATH% to administrator's
+//    HOME, so by default the xet metadata folders are isolated. If 'run as admin A' explicility configures
+//    cache or repo path to another location owned by A, ACLs for the created path inherit from the parent
+//    folder, so A still has full control.
+
+#[derive(Debug, Clone, Copy)]
 pub enum Permission {
     Regular,
     Elevated,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum FileType {
+    File,
+    Dir,
 }
 
 impl Permission {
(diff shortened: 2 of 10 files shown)


rust/gitxetcore/src/config/errors.rs
```diff
@@ -108,4 +108,7 @@ pub enum ConfigError {
 
     #[error("Error parsing local repo config file: {0}")]
     RepoConfigFileParseError(String),
+
+    #[error("Lack of read-write permission for path: {0}")]
+    LackofPermission(PathBuf),
```
hoytak, line 113: Let's call this PermissionDenied to be consistent with os nomenclature.

rust/gitxetcore/src/config/permission.rs
```diff
@@ -37,30 +54,125 @@ impl Permission {
...
-/// Checks if the program is run under elevated privilege
-
+/// Checks if the program is running under elevated privilege
 fn is_elevated() -> bool {
     // In a Unix-like environment, when a program is run with sudo,
     // the effective user ID (euid) of the process is set to 0.
     #[cfg(unix)]
-    return unsafe { libc::geteuid() == 0 };
+    {
+        unsafe { libc::getegid() == 0 }
```
hoytak, line 170: This switches to group id?

rust/gitxetcore/src/config/permission.rs
```diff
@@ -37,30 +54,125 @@ impl Permission {
...
+    /// 3. $HOME/[suggest_prefix]_[random_number]
+    /// 4. $TMPDIR/[suggest_prefix]_[random_number]
+    /// 5. $PWD/[suggest_prefix]_[random_number]
+    /// The random numbers are generated from a CSPRNG and are unlikely to collide.
+    ///
+    /// If the current process is running with elevated privileges, warn if the checked path doesn't exist and will be created.
+    ///
+    /// Return Ok(path) if the provided path is good to write into or an alternate path is suggested;
+    /// Return Err(_) if lacking permision for the provided path and unable to suggest an alternate.
+    pub fn check_or_suggest_path(
```
hoytak, line 71: I'm trying to understand a use case for an alternate directory than `$HOME/.xet/`...  If it's a random path, or some other non-standard thing, isn't it essentially the same as using a directory created through the system temporary directory?   Is this just so that everything succeeds, even if that file isn't writable?

rust/gitxetcore/src/config/permission.rs
```diff
@@ -37,30 +54,125 @@ impl Permission {
         }
     }
 
-    pub fn check_path(&self, path: &Path) {
```
hoytak, line 40: Idea for creating $HOME/.xet/ -- if we are running as root (it would be under sudo, so HOME wouldn't be changed), then we should just make the .xet/ folder match the owner and permissions from the $HOME directory.

ylow: This behavior is not the same between Linux and Mac.

ylow: On Mac $HOME doesn't change. On Linux it seems to vary between distributions.

rust/gitxetcore/src/config/permission.rs
seanses: The "effective user ID", which is set to 0 when running with 'sudo'. (Commented above)

rust/gitxetcore/src/config/permission.rs
seanses: The second search option `path_[$USER]` is to ensure a predictable path across git-xet calls. So there is a .xet folder already created with `sudo git-xet`, `git-xet` can still work and doesn't need to throw away cache, repos, or xet version check file. 

The rest three options are random trials to ensure that it always succeeds, with a slight preference to somewhere under `$HOME` because it's easier to find and (potentially) to be recovered from manually, because data transfer is billed. Same reasoning applies to constructing a predictable path name with `suggest_prefix` instead of from a pure random string.

rust/gitxetcore/src/config/permission.rs
seanses: This doesn't solve the problem that if a `.xet` is already owned by root. Also impersonating feels dangerous because... what if `.xet` as root is expected..

rust/gitxetcore/src/config/permission.rs
hoytak: Noting that it is getegid, which is the group id

rust/gitxetcore/src/config/permission.rs
hoytak: If we are running as sudo root (so it's not in root's home directory, but it's in the user's home directory), we create the .xet/ folder to inherit the permissions of the home directory.   If it's in ~root/.xet, it then creates it to be owned by root. So intentionally running as root is fine.  Plus, if you're root, then it won't matter. 

The only scenario we are worried about here is when <user> runs as sudo, causing ~user/.xet/ to be owned by root.  The scenarios where $HOME is different because it's running under sudo are irrelevant, as the regular mode xet process would only access ~user/.xet/, and these would be different. 

If we do this, then it means we just have to worry about the current cases where this has already happened, as it won't be a problem again.  I'd be fine erroring out and asking the user to run chown on that directory in that case.

rust/gitxetcore/src/config/permission.rs
hoytak: That's fair, but it could be surprising in a different way that if the user fixes their issue with ~/.xet, suddenly the cache and config values in any of these other directories would be suddenly ignored, which would effectively throw them away.

rust/gitxetcore/src/config/permission.rs
seanses: Ah good catch!

rust/gitxetcore/src/config/permission.rs
seanses: Reworked to implement this.

hoytak (approved):
LGTM!  Nice work.  Great comments.

PR #206, opened 2023-11-21
Updated test_xet_version_check.sh to be robust to API rate limits.

hoytak:
test_xet_version_check pings the github server a lot, resulting in denials due to API rate limits.  This new version uses a local file cache to reduce this query to once ever 24 hours or so, and then passes still if the query hits such API rate limits.

### DIFF
--- a/rust/gitxet/tests/integration_tests/test_xet_version_check.sh
+++ b/rust/gitxet/tests/integration_tests/test_xet_version_check.sh
@@ -7,7 +7,7 @@ global_tmp_dir=/tmp
 if [[ ! -z ${TEMP} ]] ; then
   global_tmp_dir="$TEMP"
 elif [[ ! -z ${TEMP_DIR} ]] ; then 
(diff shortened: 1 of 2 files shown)


rust/gitxet/tests/integration_tests/test_xet_version_check.sh
```diff
@@ -2,19 +2,72 @@
 set -e
 set -x
 
+global_tmp_dir=/tmp
+
+if [[ ! -z ${TEMP} ]] ; then
+  global_tmp_dir="$TEMP"
+elif [[ ! -z ${TEMP_DIR} ]] ; then 
+  global_tmp_dir="$TMP"
```
seanses, line 10: mismatch of what is checked and what is set.

hoytak: Fixed!

seanses (approved):
LGTM!

PR #208, opened 2023-11-29
Allow server only smudge on Xetea

seanses:
1. update smudge query interface to really allow server only query
2. update smudge command to self retry on mdbv1 smudge, this helps to eliminates `merkledb query` and `merkledb extract-git` on V2 repos on xetea. 

Xetea side PR: https://github.com/xetdata/xethub/pull/4770

Fix https://github.com/xetdata/xethub/issues/3838


jgodlew (approved):
LGTM! :shipit:

PR #209, opened 2023-11-29
Added query routines against shard to allow for extracting sampled chunk hashes.

hoytak:


rust/mdb_shard/src/shard_format.rs
```diff
@@ -823,6 +844,60 @@ impl MDBShardInfo {
...
     }
+
+    /// Returns a list of all the parts of this shard that have
+    pub fn read_cas_chunks_for_global_dedup<R: Read + Seek>(
+        reader: &mut R,
+        hash_filter_modulus: u64,
+    ) -> Result<Vec<MerkleHash>> {
+        let mut ret = Vec::new();
+
+        // First, go through and
```
seanses, line 855: here

PR #210, opened 2023-12-05
Fix for csv parsing skipping last line on EOF

hoytak:
Currently, the csv analyzer ignores the last line if there isn't a newline character at the end of that line.  This fixes that.


Saba9 (approved):
LGTM, thanks!

seanses (approved):
LGTM!

PR #214, opened 2023-12-07
ConfigGitPathOption::NoPath doesn't invoke git

seanses:
So running pyxet on a machine without git will not fail. Fix https://github.com/xetdata/xethub/issues/4816


hoytak (approved):
LGTM

PR #216, opened 2023-12-08
Possible fix for shard corruption issue.

hoytak:
The issue was that the IntershardReference class used the entries vector to track everything, but the num_entries field of the header was not caught up to date.  In most cases this results in entries being ignored, but in the case of the corruption I witnessed, the value in the num_entries field was much larger than the number of entries.

### DIFF
--- a/rust/mdb_shard/src/intershard_reference_structs.rs
+++ b/rust/mdb_shard/src/intershard_reference_structs.rs
@@ -191,6 +191,8 @@ impl IntershardReferenceSequence {
                 .resize(INTERSHARD_REFERENCE_SIZE_CAP, Default::default());
         }
 
+        s.metadata = IntershardReferenceSequenceHeader::new(s.entries.len());
+
         s
     }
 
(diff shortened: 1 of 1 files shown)


seanses:
I suspect the issue is due to padding. On serialization of IntershardReferenceSequenceEntry [here](https://github.com/xetdata/xet-core/blob/ac3f00c936688c1648fe619ccd5cf39324183d59/rust/mdb_shard/src/intershard_reference_structs.rs#L102) the number of bytes returned is `size_of::<IntershardReferenceSequenceEntry>()`. MerkleHash is [u64;4] and `total_dedup_chunks` is u32 so there will be 4 bytes padding returned from `size_of::<IntershardReferenceSequenceEntry>()`. But the actual number of bytes written is the size of 
`#pragma pack(1)
pub struct IntershardReferenceSequenceEntry {
    pub shard_hash: MerkleHash,
    pub total_dedup_chunks: u32,
}
`

seanses (approved):
I think

PR #217, opened 2023-12-09
Add debug hook to intercept errors raised; optionally log error propagation with backtrace. 

hoytak:
This PR adds the ability to intercept the ? with the debugger at the point of an error being raised.  In addition, if XET_LOG_EXCEPTIONS is set or it's logged to telemetry or a file, an error!() is logged with the error and the backtrace at the point of the error being converted from a system error (E.g. std::io::error) to one of the thiserror types.

To intercept the error being raised, put a breakpoint on the error_hook function. 

This requires the return type of the function to be the crate::errors::Result<_> type.  The logging and interception is is done in the From method of the conversion to the thiserror type. 

To implement this, the code contains a source drop of the thiserror crate.  This is is one commit, which doesn't need to be reviewed.  The subsequent commit adds in the intercept functionality.


seanses (commented):
Setting the env var `XET_LOG_EXCEPTIONS` doesn't seem to output exceptions, am I using it correctly?
```
di@di-mbp ~/tt/bsf13 % git-xet --version
git-xet 0.12.7-caa336f
di@di-mbp ~/tt/bsf13 % XET_LOG_LEVEL=debug XET_LOG_EXCEPTIONS=1 git-xet merkledb cas-stat
{
"total_cas_bytes" : 341349449,
"total_file_bytes" : 490366779
}
```

seanses (approved):
Tested with local Xetea, LGTM! Thanks for enabling this!

PR #218, opened 2023-12-11
Compact in-memory lookup table for chunks

hoytak:
- Implements a compact, in-memory hash table that takes a partial chunk hash and indicates which shard that chunk would be part of.  This table is loaded in the background from the chunk lookup tables embedded in the shard structure.  

- The hash table is simply an index of a 48 bit truncated hash to a 12 bit index of the shard.  This allows the lookup process to skip the linear search between shards.  In the case of collisions, all the shards are searched as before. 

- The number of chunks loaded is limited by XET_MAX_CHUNK_CACHE_SIZE, which defaults to 128M.  Each element takes 8 bytes (plus the size overhead of the hash table).   The chunks are prioritized by most recent shards first. 

- On a benchmark repo with a collection of 14 shards and about 10M chunks, this increased the cleaning filter process from 70-90 MB/s to 300-450 MB/s.


rust/mdb_shard/src/shard_file_manager.rs
```diff
@@ -129,23 +164,76 @@ impl ShardFileManager {
...
+                shards_lg.shard_list.push((s.clone(), shards_are_permanent));
+                if cur_index < u16::MAX as usize
+                    && shards_lg.chunk_lookup.len() + s.shard.total_num_chunks()
+                        < shards_lg.chunk_index_max_size
+                {
+                    for (h, (cas_start_index, cas_chunk_offset)) in s.read_all_truncated_hashes()? {
+                        if cas_chunk_offset > u16::MAX as u32 {
+                            break;
+                        }
+                        let cas_chunk_offset = cas_chunk_offset as u16;
```
seanses, line 205: Is this casting safe?

rust/mdb_shard/src/shard_file_manager.rs
```diff
@@ -129,23 +164,76 @@ impl ShardFileManager {
...
-        let mut current_lookup = self.shard_file_lookup.write().await;
-
-        for s in new_shards {
+        let mut shards_lg = self.shard_file_lookup.write().await;
+
+        // Go through and register the shards in order of newest to oldest
+        let mut new_shards: Vec<(&MDBShardFile, SystemTime)> = new_shards
+            .iter()
+            .map(|s| {
+                let modified = std::fs::metadata(&s.path)?.modified()?;
```
seanses, line 173: It's very difficult to predict which data file a user wants to change. When the loading order actually matters, i.e. when number of shards or number of chunks cross the threshold, a user may suddenly encounter degraded performance when changing a part of an old big file.
I lean towards loading in a random order. WDYT?

rust/mdb_shard/src/shard_file_manager.rs
hoytak: Given the test above, yes.  Longer ones are ignored.

rust/mdb_shard/src/shard_file_manager.rs
hoytak: It will always succeed, but since it's dedup, I think that it's more likely that content gets deduplicated against a recent shard than an older one.  Also, with the current configuration, the first 16 TB or so of chunk content is held in memory, so this only makes sense on much, much larger repos.   I'd say since it always succeeds, it's just an optimization, but most recent is more likely than other rules. .

seanses: That's fair.

seanses (approved):
LGTM!

PR #219, opened 2023-12-11
Fix #2 for shard corruption issue.

hoytak:
The previous fix for the shard issue didn't fully fix the problem; this addresses another (and hopefully final) issue.


seanses (approved):
LGTM!

PR #221, opened 2023-12-19
Fix for build breakage

hoytak:
It turns out there is a weird thing in rust's proc_macro_system that can cause the previous commit (incorporating a source drop of thiserror) to break the build.

In this case, registering a new version of thiserror may cause, on compilation, other packages (in this case opentelemetry) to use our source-dropped version of thiserror instead of the full package.  This breaks the build as our version adds a dependency on the tracing package, but this dependency is only propagated through the proc macro and doesn't seem -- for whatever reason -- to cause the compiler to make opentelemetry also depend on the tracing package.  This breaks the build, as the proc macro generates code that then has a dependency on tracing, which opentelemetry doesn't depend on. 

This PR fixes this by:
1. Renaming the local source drop to xet_error.  This way there won't be any confusion.
1. Moving the dependency on tracing out of the package.  This is done by changing the enabling code to register a callback instead of setting a flag.  The local callback then implements the tracing functionality. 

If you've encountered the above error, update to this PR.  If your build breaks, you need to really nuke things. 

1. Remove everything in the ~/.cargo/registry/ directory.
2. cargo clean
3. rebuild.


rust/xet_error/_git/logs/HEAD
```diff
@@ -1 +1 @@
-0000000000000000000000000000000000000000 e9ea67c7e251764c3c2d839b6c06d9f35b154647 Hoyt Koepke <hoytak@xethub.com> 1701979145 -0800	clone: from github.com:dtolnay/thiserror.git
+0000000000000000000000000000000000000000 e9ea67c7e251764c3c2d839b6c06d9f35b154647 Hoyt Koepke <hoytak@xethub.com> 1701979145 -0800	clone: from github.com:dtolnay/xet_error.git
```
znation, line 1: Not sure this rename makes sense?

hoytak: The issue here is that I took a source drop of the thiserror package as I needed to modify its internals to create the callback.  To effectively allow us to update thiserror in the future, I committed the .git/ folder as _git.  If we need to update the code in the future, we can rename that directory to do this effectively.  

However, when we specified thiserror locally by that name, cargo chose this one over the external version, including for all the dependent packages like opentelemetry.  This is where the error comes from, hence the rename to xet_error

znation: I see, I think we might specifically want to preserve `clone: from github.com:dtolnay/thiserror.git` in this file and not replace it with xet_error?

znation (approved):

rust/xet_error/_git/logs/HEAD
hoytak: True.  Good call.

seanses (approved):

PR #224, opened 2023-12-21
Background async iterator for parallel processing of streams and chunking. 

hoytak:
This PR splits reading, chunking, and dedup into three separate threads, causing a 2-3X speedup for the clean filter.  Benchmarks now show 500-600 MB/s on larger files on an SSD. 

This PR implements several utilities to make this process testable and simple.

trait AsyncIterator<ErrorType>:  An async iterator type that is templated over the error type and Item to replace the very specific AsyncIterator in merkledb.  This is now defined in parutils.  The async_chunk_iterator was then modified to take and conform to this iterator trait.   

AsyncDataIterator trait -- Simply a specialization of the AsyncIterator trait to use ErrorType = GitXetRepoError and Item = Vec<u8>.   This is now the type used throughout the data_processing_* classes.  In addition, the existing AsyncFileIterator now conforms to this trait. 

BufferedAsyncIterator: An async iterator implementation that is initialized using any AsyncIterator-conforming object and a buffer size and spawns a background tokio thread to keep the iterator buffer full.  At the end, the original iterator can be retrieved using the finish() method.   Additionally, this iterator can be initialized using a collection of starting objects, meaning that we don't need AsyncIteratorWithPutback any more.  

Now, with these utilities, the code to background both the data loading and the chunking is quite simple: 

```
        // load enough data from reader to determine whether this file is just pass-through or not.
        let starting_data = { ... } 
        
        // Test if the file is pass through...  If it is return. 
        ...
        
        // Now, keep loading the data on a background thread.
        let raw_data_iter =
            BufferedAsyncIterator::new_with_starting_data(starting_data, reader, Some(512)).await?;

        // Pass this into the chunker to chunk on another background thread.
        let mut generator =
            BufferedAsyncIterator::new(async_chunk_target_default(raw_data_iter), Some(4096))
                .await?;
 ```

### CODE REVIEW

File: rust/merkledb/src/async_chunk_iterator.rs [hunk 1]
```diff
@@ -402,6 +399,24 @@ lazy_static! {
     };
 }
 
+// Annoying that we have to explicitly declare this, but that is the cost of using async_trait
```
hoytak, line 402: Change to the reason being the use of Pin with a struct containing unsafe pointers.

rust/parutils/src/buffered_async_iterator.rs
```diff
@@ -0,0 +1,385 @@
...
+    /// is completed.  At that point, all items have been read into the local buffer.
+    pub fn new_detailed(
+        initial_data: Vec<It::Item>,
+        src_iter: It,
+        buffer_max_size: Option<usize>,
+    ) -> Self {
+        Self::new_impl(initial_data, src_iter, buffer_max_size, None)
+    }
+
+    /// Like new_deteailed, but also creates a tokio oneshot channel to return the src_iter object as
```
hoytak, line 75: typo

rust/merkledb/src/chunk_iterator.rs
```diff
@@ -7,6 +7,25 @@ use std::cmp::min;
 use std::io;
 use std::io::Read;
 
+/*
+// For silly reasons, we have to implement this one here.
+impl<<E: Send + Sync + 'static, A> FallibleAsyncIterator<E> for A
```
hoytak, line 12: Can be dropped -- not used anywhere.

rust/parutils/src/buffered_async_iterator.rs
```diff
@@ -0,0 +1,395 @@
...
+use tracing::info;
+
+enum BufferItem<T: Send + Sync + 'static, E: Send + Sync + 'static> {
+    Value(T),
+    Completed,
+    Error(E),
+}
+
+/// Now, create a buffered stream.  The 'static lifetime specifiers essentially require the objects to be owned
+/// objects that don't contain any non-reference components.
```
seanses, line 18: "don't contain any non-reference" or "don't contain any reference"?

rust/parutils/src/buffered_async_iterator.rs
```diff
@@ -0,0 +1,395 @@
...
+            };
+        }
+        Ok(ret)
+    }
+
+    /// If known, returns the number of items remaining.
+    fn items_remaining(&self) -> Option<usize> {
+        if self.all_items_in_buffer() {
+            let n = self.data_queue.len();
+            Some(n.saturating_sub(1))
```
seanses, line 249: A bit confused why subtracting 1 here?

rust/gitxetcore/src/lib.rs
```diff
@@ -2,8 +2,7 @@
 
 extern crate more_asserts;
 
-mod async_file_iterator;
-mod async_iterator_with_putback;
+// mod async_iterator_with_putback;
```
seanses, line 5: Let's remove this

rust/parutils/src/buffered_async_iterator.rs
seanses: Ah, it's because there's a `Complete` at the end. Let's put a comment here to explain.

seanses (approved):
LGTM! Let's fix the nits and get this in

PR #225, opened 2023-12-26
Adds in block processing for filter process

hoytak:
With this PR, chunk deduplication is queried using a block of chunks in which multiple sequential chunks may match consecutively without additional lookups.  In the case where data has already been deduplicated, this can give substantial speedups.

Depends on https://github.com/xetdata/xet-core/pull/224.


rust/gitxetcore/src/data_processing_v2.rs
```diff
@@ -131,7 +131,7 @@ impl PointerFileTranslatorV2 {
             summarydb,
             cas: cas_client,
             prefix: config.cas.prefix.clone(),
-            small_file_threshold: config.cas.size_threshold,
+            small_file_threshold: SMALL_FILE_THRESHOLD,
```
seanses, line 134: We may want to use `config.cas.size_threshold` per the changes in https://github.com/xetdata/xet-core/pull/227

seanses (approved):
One small issue above, others LGTM!

Issue #223, opened 2023-12-19
git xet clone should derive CAS endpoint information from the remote


hoytak:
Currently, the CAS data endpoint information is taken from an environment variable. If this environment variable is not present when other commits are made, then the repository can end up with the CAS data stored in multiple places, breaking it.

The CAS endpoint is purely associated with the remote, and we should derive it as needed. Configuration should be an override.

PR #227, opened 2024-01-03
Adding size threshold as a config

jgodlew:
Changes the small file threshold used in the pointer file translators from a constant to a config (e.g. `XET_CAS_SIZETHRESHOLD`). This allows applications like xet sync to treat all uploaded files as pointer files (by setting the threshold to 0) so that the sync doesn't overwhelm XetHub when uploading a dataset of thousands of moderately sized (~100KB) text files. 

We limit this config to be in the range: `[0, SMALL_FILE_THRESHOLD]` so that users don't try to upload files larger than our current pointer-file threshold.

### DIFF
--- a/rust/gitxetcore/src/config/cas.rs
+++ b/rust/gitxetcore/src/config/cas.rs
@@ -56,16 +56,15 @@ impl TryFrom<Option<&Cas>> for CasSettings {
                     None => None,
                 };
 
-                let size_threshold = match x.sizethreshold {
-                    Some(threshold) => {
-                        debug!("Cas Settings config: sizethreshold = {threshold}");
-                        if threshold > SMALL_FILE_THRESHOLD {
-                            return Err(InvalidCasSizeThreshold(threshold, SMALL_FILE_THRESHOLD))
-                        }
-                        Some(threshold)
+                let size_threshold = x.sizethreshold.map(|threshold| {
+                    debug!("Cas Settings config: sizethreshold = {threshold}");
+                    if threshold > SMALL_FILE_THRESHOLD {
+                        Err(InvalidCasSizeThreshold(threshold, SMALL_FILE_THRESHOLD))
+                    } else {
+                        Ok(threshold)
                     }
-                    None => None
-                };
+                }).transpose()?;
+
                 (endpoint, prefix, size_threshold)
             }
             None => (None, None, None),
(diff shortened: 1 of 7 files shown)


rust/gitxetcore/src/config/cas.rs
```diff
@@ -54,23 +55,37 @@ impl TryFrom<Option<&Cas>> for CasSettings {
                     }
                     None => None,
                 };
-                (endpoint, prefix)
+
+                let size_threshold = match x.sizethreshold {
```
assafvayner, line 59: low prio, more rusty way to do this

assafvayner (approved):

seanses (commented):
How do we plan to deal with this UX change: if a user xet sync with `XET_CAS_SIZETHRESHOLD = a`, then afterwards they'll have to set the same env var every time working with the repo, or else files will be cleaned differently and deemed as changes?

assafvayner:
> How do we plan to deal with this UX change: if a user xet sync with XET_CAS_SIZETHRESHOLD = a, then afterwards they'll have to set the same env var every time working with the repo, or else files will be cleaned differently and deemed as changes?

@seanses maybe we need to write this in the .xet/config.toml?

jgodlew:
Yeah, this has the same issues that the `XET_CAS_[SERVER|PREFIX]` suffers from. I agree with Assaf that having us store this in the "local" xetconfig for the repo.

hoytak:
Linking https://github.com/xetdata/xet-core/issues/223 for visibility.

seanses:
> > How do we plan to deal with this UX change: if a user xet sync with XET_CAS_SIZETHRESHOLD = a, then afterwards they'll have to set the same env var every time working with the repo, or else files will be cleaned differently and deemed as changes?
> 
> @seanses maybe we need to write this in the .xet/config.toml?

Yeah.. Another thing: We need to guarantee that pyxet never changes this value for a non-empty repo because pyxet doesn't pull files to local to change.

hoytak (approved):
Looks good to me.  I'd say get it in.  I think there are other areas where config should be put into the repo and/or be part of it.  I think some discussion on doing that in a common way would be quite useful.

rust/gitxetcore/src/config/cas.rs
jgodlew: Ok, I've replaced the match with an `Option.map(...).transpose()?` chain.

PR #228, opened 2024-01-05
GRPC error handling cleanup

hoytak:
This PR cleans up the error handling in the GRPC layer, adding Result<> to several traits to allow errors to be propagated instead of unwrapped.   

Also:
- cleaned up a few code bits along the way that could be written in a much better way.   
- Used the error.rs with a local Result as the convention.
- Fixed several clippy errors (not all immediately related to these issues, but still there).

No overall functionality should be changed, but definitely lots of small changes to the API of the internal functions.

### DIFF
--- a/rust/cache/src/lib.rs
+++ b/rust/cache/src/lib.rs
@@ -3,6 +3,7 @@
 use std::ops::Range;
 use std::{fmt::Debug, sync::Arc};
 
+use crate::error::Result;
 pub use block::BlockConverter;
 use cas::key::Key;
 use cas::singleflight;
@@ -26,8 +27,11 @@ mod xorb_cache;
 #[mockall::automock]
 #[async_trait::async_trait]
 pub trait Remote: Debug + Sync + Send {
-    async fn fetch(&self, key: &cas::key::Key, range: Range<u64>)
-        -> Result<Vec<u8>, anyhow::Error>;
+    async fn fetch(
+        &self,
+        key: &cas::key::Key,
+        range: Range<u64>,
+    ) -> std::result::Result<Vec<u8>, anyhow::Error>;
 }
 
 /// A XorbCache is the top level caching service that can be used to read
@@ -39,8 +43,8 @@ pub trait XorbCache: Debug + Sync + Send {
         key: &Key,
         range: Range<u64>,
         size: Option<u64>,
-    ) -> Result<Vec<u8>, CacheError>;
-    async fn put_cache(&self, key: &Key, contents: &[u8]) -> Result<(), CacheError>;
+    ) -> Result<Vec<u8>>;
+    async fn put_cache(&self, key: &Key, contents: &[u8]) -> Result<()>;
 }
 
 pub struct CacheConfig {
@@ -50,10 +54,10 @@ pub struct CacheConfig {
 }
 
 /// Factory method for building the XORB cache.
-pub fn from_config(
+pub fn from_config<ErrorType: Debug + Sync + Send + 'static>(
     cfg: CacheConfig,
     remote: Arc<dyn Remote>,
-) -> Result<Arc<dyn XorbCache>, CacheError> {
+) -> Result<Arc<dyn XorbCache>> {
     let cache = DiskCache::from_config(cfg.cache_dir.as_str(), cfg.capacity)?;
     let converter = BlockConverter::new(cfg.block_size);
     let request_merger = singleflight::Group::new();
--- a/rust/cache/src/xorb_cache.rs
+++ b/rust/cache/src/xorb_cache.rs
@@ -326,7 +326,7 @@ mod test {
             mock_remote
                 .expect_fetch()
                 .times(0)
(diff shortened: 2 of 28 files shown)


rust/cache/src/lib.rs
```diff
@@ -26,8 +27,11 @@ mod xorb_cache;
 #[mockall::automock]
 #[async_trait::async_trait]
 pub trait Remote: Debug + Sync + Send {
-    async fn fetch(&self, key: &cas::key::Key, range: Range<u64>)
-        -> Result<Vec<u8>, anyhow::Error>;
+    async fn fetch(
+        &self,
+        key: &cas::key::Key,
+        range: Range<u64>,
+    ) -> std::result::Result<Vec<u8>, anyhow::Error>;
```
assafvayner, line 34: slightly more concise is `-> anyhow::Error<Vec<u8>>` NBD though

rust/cache/src/xorb_cache.rs
```diff
@@ -405,7 +404,11 @@ mod test {
 
     #[async_trait::async_trait]
     impl Remote for FetchRecorder {
-        async fn fetch(&self, _key: &Key, range: Range<u64>) -> Result<Vec<u8>, Error> {
+        async fn fetch(
+            &self,
+            _key: &Key,
+            range: Range<u64>,
+        ) -> std::result::Result<Vec<u8>, anyhow::Error> {
```
assafvayner, line 411: same as above

rust/cas_client/src/caching_client.rs
```diff
@@ -77,35 +72,34 @@ impl<T: Client + Debug + Sync + Send> Client for CachingClient<T> {
...
             "CachingClient put to {}/{} of length {} bytes",
             prefix,
             hash,
             data.len()
         );
-        self.client.put(prefix, hash, data, chunk_boundaries).await
+        Ok(self
+            .client
+            .put(prefix, hash, data, chunk_boundaries)
+            .await?)
```
assafvayner, line 86: remove the `Ok` and the `?` and if the error type doesn't match use a `map_err` preferably. This looks weirdly formatted, maybe just run a rustfmt instead?

rust/cas_client/src/caching_client.rs
```diff
@@ -77,35 +72,34 @@ impl<T: Client + Debug + Sync + Send> Client for CachingClient<T> {
...
     }
 
-    async fn flush(&self) -> Result<(), CasClientError> {
+    async fn flush(&self) -> Result<()> {
         // forward flush to the underlying client
-        self.client.flush().await
+        Ok(self.client.flush().await?)
     }
```
assafvayner, line 92: same as above this is weird to me, I'd rather a map_err than unwrap and rewrap an ok

rust/cas_client/src/caching_client.rs
```diff
@@ -77,35 +72,34 @@ impl<T: Client + Debug + Sync + Send> Client for CachingClient<T> {
...
-            }
-            Ok(v) => {
-                debug!("CachingClient Get: get_length call succeeded with value {v}.");
-                v
-            }
-        };
+        let xorb_size = self.get_length(prefix, hash).await.map_err(|e| {
+            debug!("CachingClient Get: get_length reported error : {e:?}");
+            e
+        })?;
```
assafvayner, line 100: for these map_err just used for logging we should use the ErrorPrinter trait, but may have to pull it out of gitxetcore into somewhere more usable.

https://github.com/xetdata/xet-core/blob/main/rust/gitxetcore/src/log.rs#L22

assafvayner (approved):
approving to unblock, but I left some comments about changes we can do. I'd really like to see ErrorPrinter used widely instead of the map_errs for logging, but it can also wait for a future PR.

rust/cas_client/src/caching_client.rs
hoytak: In this case, I used the ? syntax intentionally as the current xet_error class now logs the error, along with a backtrace, whenever the ? operator is used to convert between Error types. map_err doesn't have the logging part, so I'd prefer to avoid it wherever possible.

rust/cas_client/src/caching_client.rs
hoytak: See comment above.

rust/cas_client/src/caching_client.rs
hoytak: I'm open to that; I think we could integrate it into the new xet_error class?

assafvayner: we could do that, the way the ErrorPrinter trait works is it's already implemented for all Result<T, E> types, so I we don't need it for the special error type, but that would be a good area od the code to put it if it's in its own crate.

rust/cas_client/src/caching_client.rs
assafvayner: might be fixable with ErrorPrinter trait then, not requirement to merge imo

PR #232, opened 2024-01-09
correct subpackage version

hoytak:
A subpackage was added with a default version instead of 0.12.8 as needed.


seanses:
Joe did the same fix in https://github.com/xetdata/xet-core/pull/231. We can close this

PR #230, opened 2024-01-08
Updated printing around the filter and pre-push process. 

hoytak:
Updated what gets printed on filter and pre-push hooks.  
- Eliminated the git-xet filter started message. 
- Moved version info to the progress printing.
- Moved constants to a common spot so places outside of gitxetcore can access constants not needed for git integration.

Now, instead of the filter printing on dedup:

```
git-xet 0.12.8 filter started
Xet: Deduplicating data blocks: 54.15 MiB | 54.15 MiB/s, done.
```

It now prints just:
```
Git-Xet 0.12.8: Processing data: 54.15 MiB | 54.15 MiB/s, done.
```

Instead of printing on download:

```
git-xet 0.12.8 filter started
Xet: Downloading data blocks: 54.15 MiB | 54.15 MiB/s, done.
```

It now prints just:
```
Git-Xet 0.12.8: Downloading data: 54.15 MiB | 54.15 MiB/s, done.
```

And on push upload, instead of:

```
"Xet: Uploading data blocks:  54.15 MiB | 54.15 MiB/s, done."
```

It prints 

```
"Git-Xet 0.12.8: Uploading data blocks:  54.15 MiB | 54.15 MiB/s, done."
```

PR #237, opened 2024-01-12
Progress reporting for downloading mdb

seanses:
Reports progress when downloading MerkleDB on repo cloning. This can take a while on large repos.
Screenshot 2024-01-12 at 5 38 00 PM

Fix https://github.com/xetdata/xethub/issues/4939

### DIFF
--- a/rust/gitxetcore/src/data_processing_v2.rs
+++ b/rust/gitxetcore/src/data_processing_v2.rs
@@ -249,7 +249,7 @@ impl PointerFileTranslatorV2 {
         if let Some(sfi) = self.shard_manager.get_shard_handle(shard_hash, false).await {
             Ok(sfi)
         } else {
-            let shard_path = download_shard(
+            let (shard_path, _) = download_shard(
                 &self.cfg,
                 &self.cas,
                 shard_hash,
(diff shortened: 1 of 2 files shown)


znation (approved):

allisonwhilden:
Always awesome to have status feedback for users. Thank you @seanses!

I have a slight question as to if we can do more to improve clarity to users in the wording. Questions I was asking myself:
- Do users understand what the MerkleDB is? Does that matter?
- Is there a difference between **Receiving** objects and **Retrieving** MerkleDB? 
- Is it common to put a version number before the colon? ("Git-Xet 0.13.0:" vs. "Git-Xet:")

These are all very minor thoughts and by no means are any reason to not merge this. Happy to just be something I can learn later. In part I'm just trying to understand some details to how we approach our CLI UX (and what are the opportunities for improvement). Thanks!

allisonwhilden (approved):

seanses:
@allisonwhilden Good questions! 
- MerkleDB may be a term too technical, how about "metadata"? In https://github.com/xetdata/xet-core/pull/230 when the actual file content is downloaded Git-Xet prints "Git-Xet 0.12.8: Downloading data", I wanted to pick a different term to show they are two steps.
- I think "Receiving objects" comes from git, that relates to downloading everything under .git/objects, which is essentially contains basic git repo information.
- A version number helps us debug when a log file is not provided, especially at the moment a client can change a lot between versions. I lean toward keeping it if not too wordy.

allisonwhilden:
Thank you! I like metadata better. Your other comments make sense to me. Thanks!

PR #238, opened 2024-01-17
git xet login check and set git user config

seanses:
Check and set "user.name" and "user.email" after successful `git xet login`.


hoytak (commented):
I'm worried about this approach as it means that when someone changes their global user name and email as git often asks one to, the global changes won't be reflected in this repo as the local config will override the global config...

erinys:
Is there a way to check if they are out of sync and ask if they want to sync?

seanses:
Close in favor of https://github.com/xetdata/xet-core/pull/242

PR #240, opened 2024-01-22
git-xet client uses HTTPS endpoint

assafvayner:
This PR should not merge before EOD PST of Tuesday 1/23/2024.

- Bump CAS Protocol version.
- Makes git-xet client's cas_client use the HTTPS data plane CAS protocol.
- processes certificate from initiate requests.

### DIFF
--- a/rust/cas_client/src/data_transport.rs
+++ b/rust/cas_client/src/data_transport.rs
@@ -117,30 +117,21 @@ impl DataTransport {
             .http2_initial_connection_window_size(HTTP2_WINDOW_SIZE)
             .http2_initial_stream_window_size(HTTP2_WINDOW_SIZE)
             .http2_only(true);
-        let connector = if let Some(root_ca) = &cas_connection_config.root_ca {
-            info!("connector with custom root ca");
-            let cert = try_from_pem(root_ca.as_bytes())?;
-            let mut root_store = rustls::RootCertStore::empty();
-            root_store.add(cert)?;
-            let config = rustls::ClientConfig::builder()
-                // add the CAS certificate to the client's root store
-                // client does not need to assume identity for authentication
-                .with_root_certificates(root_store)
-                .with_no_client_auth();
+        let root_ca = cas_connection_config.root_ca.clone().ok_or_else(|| anyhow!("missing server certificate"))?;
+        let cert = try_from_pem(root_ca.as_bytes())?;
+        let mut root_store = rustls::RootCertStore::empty();
+        root_store.add(cert)?;
+        let config = rustls::ClientConfig::builder()
+            // add the CAS certificate to the client's root store
+            // client does not need to assume identity for authentication
+            .with_root_certificates(root_store)
+            .with_no_client_auth();
 
-            HttpsConnectorBuilder::new()
+        let connector = HttpsConnectorBuilder::new()
                 .with_tls_config(config)
                 .https_only()
                 .enable_http2()
-                .build()
-        } else {
-            info!("default connector");
-            HttpsConnectorBuilder::new()
-                .with_native_roots()?
-                .https_or_http()
-                .enable_http2()
-                .build()
-        };
+                .build();
         let h2_client = builder.build(connector);
         let retry_strategy = RetryStrategy::new(NUM_RETRIES, BASE_RETRY_DELAY_MS);
         Ok(Self::new(h2_client, retry_strategy, cas_connection_config))
--- a/rust/cas_client/src/data_transport.rs
+++ b/rust/cas_client/src/data_transport.rs
@@ -372,8 +372,14 @@ impl<'a> Injector for HeaderInjector<'a> {
 #[cfg(test)]
 mod tests {
     use std::vec;
+    use lazy_static::lazy_static;
 
     use super::*;
+
+    // cert to use for testing
+    lazy_static! {
+        static ref CERT: rcgen::Certificate = rcgen::generate_simple_self_signed(vec![]).unwrap();
+    }
 
     #[tokio::test]
     async fn test_from_config() {
@@ -385,7 +391,7 @@ mod tests {
             repo_paths: "repo".to_string(),
             git_xet_version: "0.1.0".to_string(),
             root_ca: None,
-        };
+        }.with_root_ca(CERT.serialize_pem().unwrap());
         let dt = DataTransport::from_config(config).await.unwrap();
         assert_eq!(dt.authority(), endpoint);
     }
@@ -415,7 +421,7 @@ mod tests {
                 "".to_string(),
                 inner_vec.clone(),
                 "".to_string(),
-            );
+            ).with_root_ca(CERT.serialize_pem().unwrap());
             let client = DataTransport::from_config(config).await.unwrap();
             let hello = "hello".as_bytes().to_vec();
             let hash = merklehash::compute_data_hash(&hello[..]);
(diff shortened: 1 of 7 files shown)


rust/cas_client/src/data_transport.rs
```diff
@@ -54,22 +56,26 @@ enum RetryError {
...
 
     /// Should only be used for non-success errors
     #[error("Status Error: {0}")]
-    Status(http::StatusCode),
+    Status(hyper::StatusCode),
 }
 fn is_status_retriable(err: &RetryError) -> bool {
     match err {
         RetryError::Hyper(_) => true,
+        RetryError::HyperLegacy(_) => true,
```
port8080, line 72: are you sure that tls errors should be marked retryable?

rust/cas_client/src/data_transport.rs
```diff
@@ -102,18 +108,36 @@ impl DataTransport {
     /// creates the DataTransport instance for the H2 connection using
     /// CasConnectionConfig info, and additional port
     pub async fn from_config(cas_connection_config: CasConnectionConfig) -> Result<Self> {
-        debug!(
+        info!(
```
port8080, line 111: nit: make sure that you actually want this level of log in the final commit

rust/cas_client/src/data_transport.rs
```diff
@@ -102,18 +108,36 @@ impl DataTransport {
...
             .http2_initial_stream_window_size(HTTP2_WINDOW_SIZE)
-            .http2_only(true)
-            .build_http();
-
+            .http2_only(true);
+        let connector = if let Some(root_ca) = &cas_connection_config.root_ca {
+            info!("connector with custom root ca");
+            let ca = Certificate::from_pem(root_ca.as_bytes())?;
+            let tls_connector = native_tls::TlsConnector::builder()
+                .danger_accept_invalid_certs(true)
```
port8080, line 127: nit: this is such a scary function name that it deserves a comment explaining why ok

port8080 (commented):
1. Can you run the two versions with `export RUST_LOG="hyper=verbose"` and make sure that all the H2 connection negotiation params are unchanged?
2. Can you run the two versions in `rust-lldb` and ensure that the number of threads are unchanged?

rust/cas_client/src/data_transport.rs
```diff
@@ -102,18 +108,36 @@ impl DataTransport {
...
+                .add_root_certificate(ca)
+                .build()?;
+            let tls_connector = tokio_native_tls::TlsConnector::from(tls_connector);
+            let mut sub_connector = HttpConnector::new();
+            sub_connector.enforce_http(false);
+            let mut connector = HttpsConnector::from((sub_connector, tls_connector));
+            connector.https_only(true);
+            connector
+        } else {
+            info!("default connector");
```
port8080, line 137: when would this branch kick in during normal operations? Based on my reading, this is an error path

rust/cas_client/src/remote_client.rs
```diff
@@ -196,18 +220,20 @@ impl RemoteClient {
...
-        let (http_direct, grpc_direct) = self
+        let InitiateResponseEndpoints {
+            h2, put_complete
+        } = self
             .initiate_cas_server_query(prefix, hash, data.len())
             .instrument(debug_span!("remote_client.initiate"))
             .await?;
 
-        debug!("H2 Put initiate response h2 endpoint: {http_direct}, put complete endpoint {grpc_direct}");
+        debug!("H2 Put initiate response h2 endpoint: {}, put complete endpoint {}", h2.endpoint, put_complete.endpoint);
```
port8080, line 231: nit: is there any value in logging the root_ca received too?

rust/cas_client/src/data_transport.rs
jgodlew: I am having trouble understanding why we need this. Shouldn't `add_root_certificate` update the trust store with our root certificate, thereby making any requests to the CAS server valid?

rust/cas_client/src/remote_client.rs
assafvayner: Will add a log.

rust/cas_client/src/data_transport.rs
assafvayner: This would occur if the CAS server does not send the certificate in the initiate response.

In reference to: https://github.com/xetdata/xet-core/pull/240/files#diff-2047c2e47c26df755b6edd8a9a9a04e3b259758f296d6f6748c70c4855e5591cR38, this is the behavior we would want if CAS is behind this latest version, or if we set the CAS_PROTOCOL_VERSION intentionally at 0.2.0, we should still be able to handle responses lacking the root CA.

However this is an artificial requirement I added, I can treat this as an error case instead.

rust/cas_client/src/data_transport.rs
assafvayner: the HyperLegacy errors are not TLS errors, rather when the client makes a request and making the request failed. This replaces a subset of what used to be under the RetryError::Hyper(_) errors.

rust/cas_client/src/data_transport.rs
assafvayner: After some investigation I found the following:

- Setting danger_accept_invalid_certs will make encryption work regardless of specifying the certificate.
- Setting just the certificate (`add_root_certificate`) without setting the danger_accept_invalid_certs will result in a Connect error which is bizarre, telling me I'm using it wrong or it's not actually working.
  - `Error { kind: Connect, source: Some(Error { code: -67843, message: "The certificate was not trusted." }) }`

assafvayner: switching to using rustls, over native_tls, solves the issue.

rust/cas_client/src/data_transport.rs
port8080: My vote is KISS and handle it as an error

rust/cas_client/src/data_transport.rs
port8080: ack

jgodlew (approved):
LGTM! :shipit:

assafvayner:
In both cases the configuration settings logged by hyper/h2 display the same values/patterns.
```
WindowUpdate { stream_id: StreamId(0), size_increment: 5177345 }
Settings { flags: (0x0), max_concurrent_streams: 128, initial_window_size: 65536, max_frame_size: 16777215 }
Settings { flags: (0x0), max_concurrent_streams: 200, initial_window_size: 2147418112, max_frame_size: 16384, max_header_list_size: 16777216 }
Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384 }
Settings { flags: (0x0), enable_push: 0, initial_window_size: 2147418112, max_frame_size: 16384 }
```
new after changes:
```
WindowUpdate { stream_id: StreamId(0), size_increment: 5177345 }
Settings { flags: (0x0), max_concurrent_streams: 128, initial_window_size: 65536, max_frame_size: 16777215 }
Settings { flags: (0x0), max_concurrent_streams: 200, initial_window_size: 2147418112, max_frame_size: 16384, max_header_list_size: 16777216 }
Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384 }
Settings { flags: (0x0), enable_push: 0, initial_window_size: 2147418112, max_frame_size: 16384 }
```

tl;dr same exact settings on each http connection across the previous version and the new.

assafvayner:
Tracking the number of tokio tasks in flight with TLS client vs latest main (no TLS) shows a similar amount of threads smudging a 277MB file with no xorbs cached. max of around 25 tasks, with ~20 most of the runtime.

PR #241, opened 2024-01-24
Global Dedup APIs client side

seanses:
RPC definitions and shard client implementations for `ChunkProbe` and `SyncShard2`

Related issue: https://github.com/xetdata/xethub/issues/5087


rust/shard_client/src/local_shard_client.rs
```diff
@@ -1,22 +1,124 @@
...
-use std::path::PathBuf;
+use std::collections::HashMap;
+use std::io::{BufReader, Cursor, Read, Write};
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use tokio::sync::RwLock;
 
 use crate::{RegistrationClient, ShardClientInterface};
 
+struct LocalGlobalDedupTable {
```
hoytak, line 22: This name is a bit confusing...

rust/shard_client/src/local_shard_client.rs
```diff
@@ -1,22 +1,124 @@
...
+    table: Arc<RwLock<HashMap<String, String>>>, // map of prefix/chunk_hash -> prefix/shard_hash
```
hoytak, line 23: Is there ever a situation where we would have different prefixes?   If so, let's instead have a HashMap<Prefix, HashMap<MerkleHash, MerkleHash> >.  The reason here is that it would be much faster to use the regular MerkleHash as the type, as each String will require a memory allocation whereas the MerkleHash type is fixed.

hoytak: Chatted offline -- going to use a different technique here, but still use the prefix in the key and value.

rust/shard_client/src/local_shard_client.rs
seanses: Renamed to DiskBasedGlobalDedupTable and moved it to a new file global_dedup_table.rs

rust/shard_client/src/local_shard_client.rs
seanses: Finally implemented using a leveldb. Added tests for multithreaded read/write and multi-instance read/write.

hoytak (approved):
LGTM, fixed remaining issues.

PR #242, opened 2024-01-24
Updated logic around user name and user email

hoytak:
This PR updates the logic around handling the user.name and user.email fields in the config.  These were needed for adding notes, as the notes needed a signature field to be filled out.  However, as these fields are rarely visible, we now use the following logic: 
- If user.name and user.email are available from git, use those.   
- Otherwise, if the user name and email are available from the xet config, use those. 
- Finally, if none of these are available, use "Xet System" for the name and "system@xethub.com" for the backup values. 

Now, with this change, there are no longer any checks for these fields.   Tests are updated to ensure they work in this environment.  

Resolves https://github.com/xetdata/xethub/issues/5021, https://github.com/xetdata/xethub/issues/4718, https://github.com/xetdata/xethub/issues/2561.

### DIFF
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -1,4 +1,4 @@
-use crate::git_integration::{clone_xet_repo, git_user_config::verify_user_config_on_clone};
+use crate::git_integration::clone_xet_repo;
 use clap::Args;
 
 use crate::config::XetConfig;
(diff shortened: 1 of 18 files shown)


erinys (approved):
This is great!

Clarified offline that these are internal fields that don't surface anywhere, so having them set to a system default (worst case) will not affect anything. These values should be easy to modify later if needed.

rust/gitxetcore/src/command/clone.rs
```diff
@@ -27,7 +27,6 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
     let arg_v: Vec<&str> = args.arguments.iter().map(|s| s.as_ref()).collect();
 
     // First check local config.
```
seanses, line 29: Let's also remove this comment

seanses (approved):
LGTM!

PR #243, opened 2024-01-24
pull out error printer and some usages

assafvayner:
Make ErrorPrinter trait be exported out of it's own error_printer crate so that it can be imported from non-gitxetcore crates as well as projects outside of this repository.

Expand the ErrorPrinter trait to give info_error and debug_error functions so that we can log with different tracing levels.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -21,5 +21,6 @@ members = [
     "xet_error",
     "xet_config",
     "xetblob",
-    "lazy", "error_printer",
+    "lazy",
+    "error_printer",
 ]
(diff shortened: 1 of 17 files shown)


jgodlew (approved):
LGTM! :shipit:

rust/Cargo.toml
```diff
@@ -21,5 +21,5 @@ members = [
     "xet_error",
     "xet_config",
     "xetblob",
-    "lazy",
+    "lazy", "error_printer",
```
seanses, line 24: Put into a separate line?

seanses (approved):
Nice! I didn't know this crate exists.

PR #248, opened 2024-02-05
Limit the size of commits when using bbq endpoint.

hoytak:
When transaction content sizes are over 16MB, the client spits up the transactions into multiple commits, each handled separately.


hoytak:
@jgodlew -- question -- This PR does the upload of the split transactions in parallel, but not sure that's the best strategy here given concerns on the endpoint.  Should I do that sequentially?  Seems like that could take a really long time.

rust/xetblob/src/xet_repo.rs
```diff
@@ -856,41 +876,65 @@ impl XetRepoWriteTransaction {
...
+            let git_object_commit_command = serde_json::to_string(&command)
+                .map_err(|_| anyhow!("Unexpected serialization error for {:?}", command))?;
+            info!("{git_object_commit_command}");
+
+            let remote_base_url = self.remote_base_url.clone();
+
+            upload_tasks.spawn(async move {
+                perform_atomic_commit_query(remote_base_url, &git_object_commit_command).await?;
+                Ok(())
+            });
```
jgodlew, line 931: I am not sure we want to send the atomic_commit concurrently as that will just generate a bunch of load on the server. I don't expect us to get too much perf benefit here since the server can only actually commit one-by-one, so having many parallel commits just means that some of them will be forced to wait for the lock (which, notably, is an unfair lock), risking timeouts. 

True, there is some parallelism on the server-side that can speed things up (JSON parsing, verification, and preparation of the commit), which would improve performance, but I think we should test out the impact of this parallelism in a separate investigation / PR.

jgodlew (commented):
Overall, looks good, I am not really sure that parallelizing the commits is going to give us much benefit overall and we might want to examine the performance impact it will have.

jgodlew (approved):

PR #254, opened 2024-02-16
Add proper support for summarizing TSV

znation:
Pass the `\t` delimiter through wherever we can detect that the file is a TSV rather than CSV. Previously, the csv_core module assumed a `,` delimiter wherever it was used.


hoytak (approved):
LGTM

PR #258, opened 2024-02-20
rm s3 command

### LINKED ISSUE #233: s3 crate: replace use of rusoto with new aws sdk for rust

assafvayner:
fix #233.

Rather than upgrade the dependency version, no use for the s3 crate at all, better to remove it.


hoytak:
Confirming that this isn't used anywhere in our code, but LGTM.

PR #263, opened 2024-02-20
Enable global dedup query routes

hoytak:
This PR adds a global_dedup_query_policy config option alongside smudge_query_policy.   This governs when the cleaning process tries to query the shard server for dedup shards by chunk and salt.   By default, it does it for direct access methods only (E.g. xet cp).


rust/gitxetcore/src/data/remote_shard_interface.rs
```diff
@@ -110,41 +165,84 @@ impl RemoteShardInterface {
...
+            // Trigger error and backtrace
+            Err(anyhow!(
+                "ERROR: Repo salt requested but has not been configured."
+            ))?;
+            unreachable!();
+        };
+        Ok(salt)
     }
 
     pub async fn query_server(
```
seanses, line 233: This struct now supports both smudge query and dedup query so this function name becomes ambiguous. Maybe make this private and rename it?

rust/gitxetcore/src/git_integration/git_xet_repo.rs
```diff
@@ -1907,7 +1944,7 @@ impl GitXetRepo {
 
     // Add a secure random number as salt to refs notes.
     // Do nothing if a salt already exists.
-    pub fn set_repo_salt(&self) -> Result<bool> {
+    pub fn set_repo_salt(&mut self) -> Result<bool> {
```
seanses, line 1947: Do we need a `&mut`?

rust/gitxetcore/src/data/mdb.rs
```diff
@@ -458,7 +460,7 @@ fn clean_existing_v1_conversions(dir: &Path) -> errors::Result<()> {
 /// Record the converted shard in the V2 refs notes.
 /// Write a guard note in V1 refs notes.
 pub async fn upgrade_from_v1_to_v2(config: &XetConfig) -> errors::Result<()> {
-    let repo = GitXetRepo::open(config.clone())?;
+    let mut repo = GitXetRepo::open(config.clone())?;
```
seanses, line 463: `mut` not needed, also see `set_repo_salt`

rust/gitxetcore/src/constants.rs
```diff
@@ -53,3 +53,6 @@ pub const REPO_SALT_LEN: usize = 32;
 
 pub const XET_BACKUP_COMMIT_NAME: &str = "Xet System";
 pub const XET_BACKUP_COMMIT_EMAIL: &str = "system@xethub.com";
+
+// Approximately 4 MB min spacing between global dedup queries.
```
seanses, line 57: Add a comment for how this is computed: `4 MB/TARGET_CDC_CHUNK_SIZE`

seanses (approved):
LGTM!

rust/gitxetcore/src/data/remote_shard_interface.rs
hoytak: Good catch!

rust/gitxetcore/src/git_integration/git_xet_repo.rs
hoytak: Was an artifact of a previous thing.

rust/gitxetcore/src/data/remote_shard_interface.rs
hoytak: Good catch.

hoytak:
Windows runner seems to be unstable / ultra slow (other PRs too).  Merging....

PR #267, opened 2024-02-23
Enable SSCache for github actions

hoytak:
Testing...

### DIFF
--- a/rust/gitxet/build.rs
+++ b/rust/gitxet/build.rs
@@ -2,9 +2,12 @@
 use std::env;
 
 fn main() {
+    println!("cargo:rerun-if-changed=build.rs")
+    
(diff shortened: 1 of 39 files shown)


rust/gitxet/build.rs
```diff
@@ -2,6 +2,8 @@
 use std::env;
 
 fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
```
assafvayner, line 5: do we want this println?

assafvayner (approved):
maybe rm the println below. Is it working on windows? looks like the latest build still took a long time (might be worth checking if the caching is helping with another commit)

rust/gitxet/build.rs
assafvayner: nevermind, I just found out you need to print this line to compile on change only, a bit surprised it's a println not another macro.

PR #276, opened 2024-02-29
Tableau Workbook Summary

jgodlew:
Adds Tableau Workbook Summarization to git-xet, allowing us to summarize `.twb` files.
The summarization works is done in the following high-level steps:
1. As blocks are cleaned from the `.twb` file, they're added to a buffer by the `TwbAnalyzer`
2. Once complete, the `TwbAnalyzer::finalize()` method is called, which kicks off the summarization.
3. The buffered twb file is converted to a raw XML tree (using roxmltree crate).
4. The XML node tree is then converted into a `TwbRaw` struct, containing a series of "raw" structs that roughly map to the XML nodes in the file.
5. We then process the `TwbRaw` struct into a `TwbSummary` struct, which provides the semantic concepts that the XetHub UI / Users expect (e.g. logical tables and display names).
6. This `TwbSummary` is then returned back to the filter logic, where it can be serialized and stored in the summary_db for the repo.

The layout of the `tableau_summary` crate follows fairly closely to the above structure with a module for the twb summarization and sub-modules for the `raw` and `summary` passes.

Note: there is likely a more optimal processing of the file in a streaming manner, however, this is much more complex than simply going through 3 fairly isolated passes. If performance becomes a problem, we can optimize later.


rust/tableau_summary/src/twb/raw/datasource/connection.rs
```diff
@@ -0,0 +1,489 @@
...
+        let tables = n.find_tagged_children("relation")
+            .into_iter()
+            .map(Table::from)
+            .map(|table| (table.name.clone(), table))
+            .collect();
+        let clause = n.get_tagged_child("clause")
+            .map(|c| Clause {
+                clause_type: c.get_attr("type"),
+                expression: c.get_tagged_child("expression").map(Expression::from).unwrap_or_default(),
+            }).unwrap_or_default();
```
assafvayner, line 228: mgiht be worth a Clause::from(whatever c is, Node I'm guessing)

rust/tableau_summary/src/twb/raw/datasource/connection.rs
```diff
@@ -0,0 +1,489 @@
...
+        info!("found col value: {val} that didn't parse into 2 parts");
+        return None;
+    };
+    Some((key, val_tuple))
+}
+
+// Given a string like `[foo].[bar]` return ["foo", "bar"]
+// TODO: see what tds / twb files do when the table or column has a `.` in the name
+//       currently, we just assume it is not escaped or anything
+//       (e.g. "[foo.1].[bar.txt]" -> ["foo.1", "bar.txt"])
```
assafvayner, line 359: sadly this sounds like a good regex problem.
starting point: https://chat.openai.com/share/8b8a54e3-58be-4585-b848-7cde9a964363

rust/tableau_summary/src/twb/raw/datasource/connection.rs
```diff
@@ -0,0 +1,489 @@
...
+    fn from(n: Node) -> Self {
+        if n.get_tag() != "metadata-records" {
+            info!("trying to convert a ({}) to metadata-records", n.get_tag());
+            return Self::default();
+        }
+        let mut capabilities = HashMap::new();
+        let mut columns = HashMap::new();
+        n.find_tagged_children("metadata-record")
+            .into_iter()
+            .for_each(|c| {
```
assafvayner, line 407: `for c in n.find_tagged_children("metadata-record") {`
you can deindent a bit and also don't need a separate function call per item (although I assume this is inlined by the compiler)

rust/tableau_summary/src/twb/raw/datasource/substituter.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Given the logical source name and the column name, get the display name for the
+    /// column.
+    fn find_column_for_source(&self, source: &str, name: &str) -> Option<Cow<str>>;
+}
+
+impl<'a, T: ColumnFinder> Substituter<'a, T> {
+    /// Given the parameterized string, replace any referenced columns with their Caption'ed representation
+    /// e.g. if the column: `[Calc_12345]` has the caption: `'Orders made'`, then given the string:
+    /// `"CEIL([Calc_12345])"` we will output: `"CEIL([Orders made])"`.
+    /// This will also return a list of dependencies found in the string: at tuple of (datasource, column)
```
assafvayner, line 21: this again feels a bit like a problem to be solved with a regex

assafvayner: if not, at least create more helper functions for each case below, because it's quite a lot to parse at once.

rust/tableau_summary/src/twb/raw/worksheet/table.rs
```diff
@@ -0,0 +1,218 @@
...
+        if n.get_tag() != "view" && n.get_tag() != "dashboard" {
+            info!("trying to convert a ({}) to a worksheet view or dashboard view", n.get_tag());
+            return Self::default();
+        }
+
+        let sources = n.find_tagged_children("datasource-dependencies")
+            .iter().map(|n| (n.get_attr("datasource"), *n))
+            .map(|(name, node)| {
+                (name, get_column_dep_map(node))
+            }).collect();
```
assafvayner, line 66:
```rust
 .iter()
    .map(|n| (n.get_attr("datasource"), get_column_dep_map(*n)))
    .collect();
```

rust/tableau_summary/src/twb/summary/datasource.rs
```diff
@@ -0,0 +1,298 @@
...
+        });
+
+    // try to add any unchanged columns not found in the column_set (i.e. those in metadata)
+    let m = datasource.connection
+        .as_ref()
+        .map(|c| &c.metadata_records.columns)
+        .unwrap_or(&HashMap::new())
+        .iter()
+        .filter(|(k, _)| !columns.contains_key(*k))
+        .filter(|(k, _)| !drill_columns.contains_key(*k))
```
assafvayner, line 151: can be combined into 1 expression

assafvayner (commented):
leaving comments, I think final approval should go to @sirahd

rust/tableau_summary/src/twb/raw/worksheet/table.rs
```diff
@@ -0,0 +1,218 @@
...
+            ColumnDep::ColumnInstance(ci) => Cow::from(self.column_instance_caption(ci)),
+            ColumnDep::Group(g) => get_captioned(g.caption.as_str(), g.name.as_str()),
+            ColumnDep::Table(_) => Cow::from("")
+        }
+    }
+
+    pub fn get_column(&self, source: &str, name: &str) -> Option<&ColumnDep> {
+        if source.is_empty() {
+            self.sources.values()
+                .filter_map(|m|m.get(name))
```
sirahd, line 102: [nit]
```suggestion
                .filter_map(|m| m.get(name))
```

rust/tableau_summary/src/twb/summary/datasource.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
+pub struct Table {
+    name: String,
+    dimensions: Vec<Column>,
+    measures: Vec<Column>,
+}
+
+// TODO: can't skip serialization or else bincode serialization (used for db file) will
```
sirahd, line 23: [nit] remove this TODO?

rust/tableau_summary/src/xml.rs
```diff
@@ -0,0 +1,174 @@
...
+            return v;
+        }
+
+        v.extend(self.get_children()
+            .flat_map(|ch| ch.find_tagged_decendants_to_depth(tag_name, depth - 1)));
+        v
+    }
+
+    /// Gets all decendant nodes with the specified tag
+    fn find_all_tagged_decendants(&self, tag_name: &str) -> Vec<Self> {
```
sirahd, line 37: [nit]
```suggestion
    fn find_all_tagged_descendants(&self, tag_name: &str) -> Vec<Self> {
```

rust/tableau_summary/src/xml.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    fn get_attr(&self, attr: &str) -> String {
+        self.get_maybe_attr(attr).unwrap_or_default()
+    }
+    
+    fn get_text(&self) -> String;
+
+    /// Depth is how many generations we should dive down to:
+    /// 0 == check this node, 1 == children, 2 == grand-children, ...
+    fn find_tagged_decendants_to_depth(&self, tag_name: &str, depth: i8) -> Vec<Self> {
```
sirahd, line 20: [nit]
```suggestion
    fn find_tagged_descendants_to_depth(&self, tag_name: &str, depth: i8) -> Vec<Self> {
```

sirahd (approved):
LGTM! :shipit:

rust/tableau_summary/src/twb/printer.rs
```diff
@@ -0,0 +1,40 @@
...
+    println!("{json}");
+    Ok(())
+}
+
+// Reads the whole file from disk, and returns the Twb analysis.
+// Intended to be used for small passthrough (non-pointer) files.
+pub fn summarize_twb_from_reader(file: &mut impl Read) -> anyhow::Result<Option<TwbSummary>> {
+    let mut analyzer = TwbAnalyzer::default();
+
+    let mut chunk: Vec<u8> = vec![0; 65536];
```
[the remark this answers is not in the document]
jgodlew: It is just a size for the length of the buffer. This is similar to the code used by the csv summary, so I just re-used that. I can extract it to a constant.

rust/tableau_summary/src/twb/raw/datasource.rs
```diff
@@ -0,0 +1,140 @@
...
+impl ColumnFinder for RawDatasource {
+
+    fn find_column(&self, name: &str) -> Option<Cow<str>> {
+        self.column_set.columns.get(name)
+            .and_then(ColumnDep::get_column)
+            .map(get_column_captioned)
+    }
+
+    fn find_column_for_source(&self, source: &str, name: &str) -> Option<Cow<str>> {
+        if let Some(dep) = self.dependencies.get(strip_brackets(source)) {
```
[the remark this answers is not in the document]
jgodlew: yeah, I forgot that the Try operator works with Options too.

rust/tableau_summary/src/twb/raw/datasource/object_graph.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub relationships: Vec<Relationship>,
+}
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
+pub struct TableauObject {
+    pub id: String,
+    pub caption: String,
+    pub table: Option<Table>,
+    pub join: Option<Join>,
+    pub union: Option<Union>,
```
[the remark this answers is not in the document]
jgodlew: Yeah, actually, I can just use the Relation enum that already exists.

rust/tableau_summary/src/twb/raw/datasource/columns.rs
```diff
@@ -0,0 +1,252 @@
...
+
+impl<'a, 'b> TryFrom<Node<'a, 'b>> for ColumnDep {
+    type Error = ();
+
+    fn try_from(n: Node) -> Result<Self, Self::Error> {
+        Ok(match n.get_tag() {
+            "column" => Column(n.into()),
+            "column-instance" => ColumnInstance(n.into()),
+            "group" => Group(n.into()),
+            "_.fcp.ObjectModelTableType.true...column" => Table(n.into()),
```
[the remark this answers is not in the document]
jgodlew: added comment to the constant

rust/tableau_summary/src/twb/raw/datasource.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.dependencies.values_mut().for_each(|dep| {
+            dep.caption = captions.get(&dep.name).cloned().unwrap_or_default();
+        });
+    }
+}
+
+/// returns the caption for the column surrounded by `[]`.
+fn get_column_captioned(c: &ColumnMeta) -> Cow<str> {
+    if c.caption.is_empty() {
+        Cow::from(c.name.as_str())
```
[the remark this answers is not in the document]
jgodlew: We need a Cow here because we might be returning either an &str (reference to the name) or a String (bracketed caption). We can't return an &str since the bracketed caption was created within this method and we would need to transfer ownership of that String to the caller. 
We could return a String (owned) instead. However, that requires cloning the name reference, since we can't take ownership of `c.name` (as c is a reference to a ColumnMeta). Cloning would waste memory, especially in a commonly used method like this where we only really need a string reference.

Cow<str> provides us with the way to provide either an owned String or an &str and won't clone the &str unless the caller needs to. So, overall, this is a bit of a memory optimization for the code.

rust/tableau_summary/src/twb/raw/datasource/substituter.rs
jgodlew: Yeah, the problem is 3 fold:
1. We need to do a substitution with extra functionality (need to lookup the column name for substitution). It isn't just substitute "foo" for "bar". This might be doable with the regex crate.
2. The substitution needs to be made in two different use-cases (`[datasource].[column]` or `[column]`). This could be possible, but not without a significant chunk of work to figure out what that regex would look like and how it can work in conjunction with Issue1.
3. Lastly, we are not just substituting, we are also capturing as well. AFAICT, the regex crate doesn't support this.

I've refactored the Substituter to break up the logic better and "hopefully" make it easier to understand.

rust/tableau_summary/src/twb/raw/datasource/columns.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+}
+
+impl<'a, 'b> From<Node<'a, 'b>> for GroupMeta {
+    fn from(n: Node) -> Self {
+        if n.get_tag() != "group" {
+            info!("trying to convert a ({}) to group", n.get_tag());
+            return Self::default();
+        }
```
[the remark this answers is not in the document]
jgodlew: Yeah, these have popped up pretty frequently. I've added a simple macro to help with this.

PR #280, opened 2024-02-29
WSL faster for windows testing?

hoytak:
Hypothesis: yes it is...


hoytak:
Holding off on this... maybe not worth it?

PR #291, opened 2024-03-08
Add threshold to csv summary

seanses:
Fix https://github.com/xetdata/xethub/issues/5559

With this PR by default a CSV summary will not be created for files under 256 KiB. This can be tuned by `export XET_CSV_MIN_SIZE = [file size in bytes]`


hoytak (approved):
LGTM -- ideally, need this in the config, but for our purposes this works fine.

seanses:
> LGTM -- ideally, need this in the config, but for our purposes this works fine.

Given there's a plan to move summary analyzer out of git-xet to actions I think we can just use this temporary fix

PR #293, opened 2024-03-12
prelude to compression on the wire for dataplane

assafvayner:
Adds header on client to accept encoding, though set to none at the moment until server side implementation can support and respect this header.

This PR is needed so that the CAS server can use utils::CompressionScheme and share the header values.

### DIFF
--- a/rust/cas_client/src/data_transport.rs
+++ b/rust/cas_client/src/data_transport.rs
@@ -204,9 +204,9 @@ impl DataTransport {
             let cas_accept_encoding_value = HeaderValue::from_static(Into::into(CompressionScheme::None));
             req = req.header(cas_accept_encoding_header, cas_accept_encoding_value);
         } else if method == Method::POST {
-            let cas_accept_encoding_header = HeaderName::from_static(CAS_CONTENT_ENCODING_HEADER);
-            let cas_accept_encoding_value = HeaderValue::from_static(Into::into(CompressionScheme::None));
-            req = req.header(cas_accept_encoding_header, cas_accept_encoding_value);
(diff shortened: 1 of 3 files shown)


rust/utils/src/compression.rs
```diff
@@ -0,0 +1,33 @@
+use std::str::FromStr;
+use anyhow::anyhow;
+
+pub const CAS_CONTENT_ENCODING_HEADER: &str = "xet-cas-content-encoding";
```
seanses, line 4: Where to use this value?

assafvayner: This will be used in 2 spots:
- on puts with compression from the client
- on get responses from the server. client will use this header to determine the encoding.

We will not be using the standard HTTP Content-Encoding header due to the fact that LZ4 is not one of the "official" supported encodings.

rust/cas_client/src/data_transport.rs
```diff
@@ -197,6 +198,17 @@ impl DataTransport {
             .header(cas_protocol_version_header, cas_protocol_version)
             .uri(&dest)
             .version(Version::HTTP_2);
+
+        if method == Method::GET {
+            let cas_accept_encoding_header = HeaderName::from_static(CAS_ACCEPT_ENCODING_HEADER);
+            let cas_accept_encoding_value = HeaderValue::from_static(Into::into(CompressionScheme::None));
+            req = req.header(cas_accept_encoding_header, cas_accept_encoding_value);
+        } else if method == Method::POST {
+            let cas_accept_encoding_header = HeaderName::from_static(CAS_CONTENT_ENCODING_HEADER);
```
seanses, line 207: maybe rename to `cas_content_encoding_header` to match the constant name.

seanses (approved):
See one nit above, others LGTM

PR #294, opened 2024-03-12
Enable git-xet login with s3 credentials

ylow:
This PR enables the reading and writing of xethub S3 credentials to xet config.

Running "git-xet login" without the s3 option will print the following
```
Set the following environment variables to use AWS cli
Re-run `git-xet login` with the --s3 option to write these keys to an AWS cli configuration profile

	AWS_ACCESS_KEY_ID = blah
	AWS_SECRET_ACCESS_KEY = blah

To see the repos you have access to via S3,
set the environment variables above and run:
	aws --endpoint-url=s3.xethub.com s3 ls
```

Also, adds the --s3 option will try to write the config and will print the following message:
```
The following was config written successfully to AWS cli configuration.

	AWS_ACCESS_KEY_ID = blah
	AWS_SECRET_ACCESS_KEY = blah

To see the repos you have access to via S3, run:
	aws --endpoint-url=s3.xethub.com --profile xethub s3 ls
```

If --s3 is provided and aws-cli failed to run , it will print
```
Failed to run aws cli.
Unable to configure S3 automatically.
```
and will fall back to the none --s3 printout. (It will also not print the Re-run .... line)

### DIFF
--- a/rust/gitxetcore/src/command/login.rs
+++ b/rust/gitxetcore/src/command/login.rs
@@ -3,6 +3,7 @@ use crate::config::{get_global_config, XetConfig};
 use crate::errors;
 use anyhow::anyhow;
 use clap::Args;
+use std::process::Command;
 use tracing::{error, warn};
 use xet_config::{Axe, Cas, Cfg, User};
 
@@ -31,6 +32,10 @@ pub struct LoginArgs {
     /// Do not overwrite credentials if they already exist
     #[clap(long)]
     pub no_overwrite: bool,
+
+    /// Configures aws-cli for xethub S3 access
+    #[clap(long)]
+    pub s3: bool,
 }
 
 /// applies config from LoginArgs onto a cfg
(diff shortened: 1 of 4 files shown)


rust/gitxetcore/src/command/login.rs
```diff
@@ -82,10 +87,167 @@ fn apply_config(
...
+    let aws_secret_key = user
+        .aws_secret_key
+        .as_ref()
+        .ok_or(errors::GitXetRepoError::AuthError(anyhow!(
+            "Not authenticated. Run git-xet login first"
+        )))?;
+    // dots and dashes are ok
+    // this makes the profile name slightly more readable
+    let mut profilename = host.to_string();
+    profilename.retain(|x| x.is_alphanumeric() || x == '.' || x == '-');
```
hoytak, line 128: _ ?

rust/gitxetcore/src/command/login.rs
```diff
@@ -82,10 +87,167 @@ fn apply_config(
...
+/// Prints the AWS Config stdout.
+/// Does not do anything if there is no user config.
+fn print_s3_config(cfg: &Cfg) -> errors::Result<()> {
+    if let Some(ref user) = cfg.user {
+        let aws_access_key =
+            &user
+                .aws_access_key
+                .as_ref()
+                .ok_or(errors::GitXetRepoError::AuthError(anyhow!(
+                    "Not authenticated. Run git-xet login first"
```
hoytak, line 175: IDK if it matters, but with all the error messages being the same, we won't know which line caused an error if a user reports it.

rust/gitxetcore/src/command/login.rs
```diff
@@ -82,10 +87,167 @@ fn apply_config(
...
+        ok = false;
+    }
+    if !ok {
+        // instructions without an aws profile
+        eprintln!("Set the following environment variables to use AWS cli");
+        if !failed_to_run_awscli {
+            eprintln!("Re-run `git-xet login` with the --s3 option to write these keys to an AWS cli configuration profile")
+        }
+        eprintln!("");
+        let _ = print_s3_config(cfg);
```
hoytak, line 232: Trying to trace out the path when things are not set up well and it seems like the error could stack up on each other without making sense:  E.g.:

Failed to run aws cli.
Unable to configure S3 automatically.
Not authenticated. Run git-xet login first.
To see the repos you have access to via S3,
set the environment variables above and run:
    aws --endpoint-url=s3.xethub.com s3 ls

hoytak (approved):
LGTM, minor comments.

rust/gitxetcore/src/command/login.rs
ylow: The errors are all eaten and not printed. (let _ = print_s3_config)

rust/gitxetcore/src/command/login.rs
ylow: I don't actually know! I just manually tested . and -

rust/gitxetcore/src/command/login.rs
hoytak: These are the eprintln! statements, not the errors.

PR #297, opened 2024-03-14
Update to allow controlled iteration on a git tree.

hoytak:


hoytak:
Closed in favor of https://github.com/xetdata/xet-core/pull/302.

PR #298, opened 2024-03-14
compression refactor to protobuf type

assafvayner:
Refactor CompressionScheme type to be defined in protobuf so the CAS server will report accepted encodings for put operation.

### DIFF
--- a/rust/cas_client/src/data_transport.rs
+++ b/rust/cas_client/src/data_transport.rs
@@ -25,7 +25,8 @@ use tokio_rustls::rustls;
 use tokio_rustls::rustls::pki_types::CertificateDer;
 use tracing::{debug, error, info, info_span, warn, Instrument, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
-use cas::compression::{CAS_ACCEPT_ENCODING_HEADER, CAS_CONTENT_ENCODING_HEADER, CompressionScheme};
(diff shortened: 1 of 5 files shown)


jgodlew (approved):
LGTM! :shipit:

rust/cas_client/src/data_transport.rs
```diff
@@ -335,9 +328,13 @@ impl DataTransport {
...
             .retry(
                 || async {
-                    let req = self
+                    let full_size = data.len();
+                    // compression of data to be done here, for now none.
+                    let mut req = self
                         .setup_request(Method::POST, prefix, hash, Some(data.to_owned()))
                         .map_err(RetryError::from)?;
+                    let headers = req.headers_mut();
+                    headers.insert(CAS_INFLATED_SIZE_HEADER, HeaderValue::from(full_size));
```
seanses, line 337: Is this for verification of content on the cas server?

seanses (approved):

rust/cas_client/src/data_transport.rs
assafvayner: We need the total inflated size for decompression. We can either supply it ourselves like this, or alternatively we can  pack it into the compressed content, but according to YL this could block us from doing decompression while streaming. So we'll use this instead.

PR #299, opened 2024-03-14
strip trailing slash from endpoint parsing instead of leading slash

ylow:


rust/gitxetcore/src/git_integration/git_url.rs
```diff
@@ -79,7 +79,7 @@ impl XetPathInfo {
     /// The logic is mostly borrowed from pyxet.
     fn parse(url: &str, force_domain: &str) -> Result<Self> {
         debug!("Parsing URL '{url}', force_domain = '{force_domain}'");
-        let url = url.strip_prefix('/').unwrap_or(url);
+        let url = url.strip_suffix('/').unwrap_or(url);
```
seanses, line 82: The cp logic is implemented to respect the trailing slash: https://filesystem-spec.readthedocs.io/en/latest/copying.html#:~:text=If%20the%20trailing%20slash%20is%20omitted%20from%20%22source/subdir%22%20then%20the%20subdir%20is%20also%20copied%2C%20not%20just%20its%20contents%3A. Removing the trailing slash will lead to test failures.

seanses: Should instead just fix the port parsing part.

ylow: oh dang. ok.

hoytak:
No longer needed with recent work.

PR #301, opened 2024-03-15
Make git-xet cp support wildcard

sirahd:


hoytak (changes requested):
A couple issues, some commented on.  Also, let's add some tests for this.

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -63,80 +63,102 @@ async fn build_cp_operation_list(
...
                 return Err(GitXetRepoError::InvalidOperation(format!(
-                    "Source path {src} does not exist as a directory."
+                    "Invalid glob {src}. Wildcards can only appear in the last position",
                 )));
             }
-
+            src_path = src_root_dir.unwrap_or("").to_string();
             src_is_directory = true;
+            dest_path = dest_base_path.to_owned();
+            dest_dir = dest_path.clone();
```
hoytak, line 95: I think src_size should be set?

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -63,80 +63,102 @@ async fn build_cp_operation_list(
...
+            // we only accept globs of the for blah/blah/blah/[glob]
+            // i.e. the glob is only in the last component
+            // src_root_dir should be blah/blah/blah here
+            let src_root_dir = src_fs.parent(src.as_str());
+            
+            if let Some(src_root_dir_path) = src_root_dir {
+                if src_root_dir_path.contains('*') {}
+            }
+            
+            if src_root_dir.is_some_and(|d| d.contains('*'))   {
```
hoytak, line 87: These tests just look for the existence of `*`; without implementing glob-style filters, we should test that it's exactly `*`.  Currently, it means that `*.jpg` would behave the same as `*`.

We can either implement any glob using a filter (as in pyxet), or just error out if the last component is not exactly `*` However, in that case, the only real use case would be copying the file contents of a directory non-recursively.  Is that enough?

seanses:
*nix shells generally expand "*" to a complete file name list before sending them to the program. Is this for Windows?

sirahd:
> *nix shells generally expand "*" to a complete file name list before sending them to the program. Is this for Windows?

No, this is to replicate the same behavior that `xet cp` command offers in `git-xet cp` mainly for Tableau integration stuff.

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
sirahd: I didn't see pyxet implementation setting the `src_size` (via `src_info` in https://github.com/xetdata/pyxet/blob/c9a6a12fddbffaa32387dcc02c4e39156f895efe/python/pyxet/pyxet/file_operations.py#L192), so I didn't explicitly set the `src_size` in the wildcard case. Let me know if I miss anything important here.

hoytak: That may be a bug then.  pyxet is quite dirt road

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -63,80 +66,104 @@ async fn build_cp_operation_list(
...
-
-        if src.contains('*') {
-            return Err(GitXetRepoError::InvalidOperation(
-                "Error: Currently, wildcards are not supported.".to_owned(),
-            ));
-        }
+        let dest_dir: String;
+        let dest_path: String;
+        let mut src_size = 0;
+        let mut file_filter_pattern: String = "".to_string();
```
hoytak, line 72: Use Some/None here.

let mut file_filter_pattern = None

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -63,80 +66,104 @@ async fn build_cp_operation_list(
...
+        if src.contains('*') {
+            // Handling wildcard cases
+            // The file card matching works by recasting the operation as a directory -> directory copy,
+            // but with a possible filter pattern.
+
+            // validate
+            // we only accept globs of the for blah/blah/blah/[glob]
+            // i.e. the glob is only in the last component
+            // src_root_dir should be blah/blah/blah here
+            let src_root_dir = src_fs.parent(src.as_str());
```
hoytak, line 85: Go ahead and set 
`let src_file_name = src_fs.file_name(src.as_str());`

Use this below.

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -63,80 +66,104 @@ async fn build_cp_operation_list(
...
-                    "Source path {src} does not exist as a directory."
+                    "Invalid glob {src}. Wildcards can only appear in the last position",
                 )));
             }
-
+            src_path = src_root_dir.unwrap_or("").to_string();
             src_is_directory = true;
+            dest_path = dest_base_path.to_owned();
+            dest_dir = dest_path.clone();
+            file_filter_pattern = if !src_fs.file_name(src.as_str()).is_empty() {
```
hoytak, line 96: Set 
`file_filter_pattern = Some(Pattern::new(file_filter_pattern.as_str()).map_err(|e| GitXetRepoError::InvalidOperation(format!("Error in glob pattern {src_file_name}: {e:?})))`?;

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -155,7 +182,16 @@ async fn build_cp_operation_list(
                         size: 0,
                     });
                 } else {
-                    cp_ops.extend(src_fs.listdir(&src_path, true).await?.into_iter().map(
+                    cp_ops.extend(src_fs.listdir(&src_path, true).await?.into_iter().
+                        // filter files under dir that do not match glob pattern if the input src uses wildcard
+                        filter(|entry| {
+                            if file_filter_pattern.is_empty() {
```
hoytak, line 188: Use the 
```
if let Some(filter) = file_filter_pattern.as_ref() {
    matches(&entry.name) } else { true} 
```

rust/gitxetcore/src/xetblob/file_operations/cp_operations.rs
```diff
@@ -501,4 +540,55 @@ mod tests {
         assert!(copied_file_path.exists());
         assert_eq!(fs::read_to_string(copied_file_path).unwrap(), "Hello");
     }
+
+    #[tokio::test]
+    async fn test_wildcard_file_to_nonexistent_directory_destination() {
```
hoytak, line 545: The test logic we need to make sure works is cp my_dir/* to existing_dir/, as branches must exist in this method.    Just make sure we have a test covering that.

hoytak (approved):
LGTM for now.  Ideally we would support the non-recursive case too instead of just erroring out, but how about we file an issue for that and get this in.

PR #303, opened 2024-03-18
client work cas compression in transit

assafvayner:
client side work adding LZ4 compression in transit on the CAS H2 dataplane.

- on the download path, client advertises acceptable formats via an H2 request header
- on upload path, CAS server will reply with acceptable formats in gRPC InitiateResponse


jgodlew (approved):
LGTM! :shipit:

PR #304, opened 2024-03-18
Adding summary for TDS files

jgodlew:
Adds summaries for Tableau Datasource files (`*.tds`). This format is pretty much the same as the datasource section of the `twb` files and re-uses most of the parsing and summary code for those.


rust/libmagic/src/file_types.rs
```diff
@@ -651,8 +651,8 @@ static FILE_TYPES: phf::Map<&'static str, FileTypeInfo> = phf_map! {
     "tbm" => FileTypeInfo { friendly_type: "Tableau Bookmark", mime_type: "application/octet-stream"},
     "tcap" => FileTypeInfo { friendly_type: "3rd Generation Partnership Project - Transaction Capabilities Application Part", mime_type: "application/vnd.3gpp2.tcap"},
     "tcl" => FileTypeInfo { friendly_type: "Tcl Script", mime_type: "application/x-tcl"},
-    "tds" => FileTypeInfo { friendly_type: "Tableau Data Source", mime_type: "application/octet-stream"},
-    "tdsx" => FileTypeInfo { friendly_type: "Tableau Data Source", mime_type: "application/octet-stream"},
+    "tds" => FileTypeInfo { friendly_type: "Tableau Data Source", mime_type: "application/tds"},
```
znation, line 654: Are these mime types commonly used? I don't see them listed in https://www.iana.org/assignments/media-types/media-types.xhtml

znation (approved):

rust/gitxetcore/src/summaries/analysis.rs
```diff
@@ -27,6 +29,9 @@ impl FileAnalyzers {
         if let Some(twb) = &mut self.twb {
             twb.process_chunk(chunk);
         }
+        if let Some(tds) = &mut self.tds {
+            tds.process_chunk(chunk);
+        }
```
assafvayner, line 34: if you need to add one more ever, then maybe we make this be dynamic dispatch with an more OOP style.

```rust
pub trait FileAnalyzer {
  fn process_chunk(&self, chunk: &[u8]) -> Result<()>
}

pub struct FileAnalyers {
  analyzers: Vec<Box<dyn FileAnalyzer>>,
}

impl FileAnalyzers {
  fn process_chunk_impl(&mut self, chunk: &[u8]) -> Result<()> {
    for analyzer in self.analyzers {
      analyzer.process_chunk(chunk)?;
    }
  }
}
```

assafvayner (approved):

rust/libmagic/src/file_types.rs
jgodlew: I am not sure if they're commonly used. My understanding is that the mime-types used in here are being used by the summary code ([fetcher.rs#191](https://github.com/xetdata/xet-core/pull/304/files#diff-8bbee25fdd1cbb41082b2a710f90b0a3c5b4754fb0af2c4483f498e95b03c254R191-R200)) to select a type of summary. Using octet-stream is fairly broad and could result in us trying to generate tds summaries for many arbitrary binary types.

rust/gitxetcore/src/summaries/analysis.rs
jgodlew: Yeah, we could probably refactor this if we start to get more summaries, but considering we are moving summaries to custom-views v2 / XetMeta, likely in the next sprint, it isn't too important to refactor these implementations.

rust/libmagic/src/file_types.rs
znation: Fair enough -- you're right that the logic there is affected by this mime type, so I agree let's use a specific type. LGTM.

PR #305, opened 2024-03-19
Fix Summary Schema

jgodlew:
Moves the Tableau summaries to a versioned `SummaryExt` struct that will hold additional summaries. This seems to be working locally for csv files summarized earlier and seems to work for tableau files too.

### DIFF
--- a/rust/gitxetcore/src/command/summary.rs
+++ b/rust/gitxetcore/src/command/summary.rs
@@ -8,7 +8,9 @@ use std::{
 };
 use tracing::warn;
 use tableau_summary::tds::printer::{print_tds_summary, print_tds_summary_from_reader};
+use tableau_summary::tds::TdsSummary;
 use tableau_summary::twb::printer::{print_twb_summary, print_twb_summary_from_reader};
+use tableau_summary::twb::TwbSummary;
 
 use crate::{config::XetConfig, errors::GitXetRepoError, utils};
 use crate::{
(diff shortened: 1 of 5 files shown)


rust/gitxetcore/src/command/summary.rs
```diff
@@ -103,8 +105,20 @@ async fn print_summary_from_db(
     match summary_type {
         SummaryType::Libmagic => print_stored_summary_impl(&summary.libmagic),
         SummaryType::Csv => print_stored_summary_impl(&summary.csv),
-        SummaryType::Twb => print_stored_summary_impl(&summary.twb),
-        SummaryType::Tds => print_stored_summary_impl(&summary.tds),
+        SummaryType::Twb => {
+            if let Some(sum) = summary.additional_summaries.as_ref() {
```
hoytak, line 109: nit:  Could do `print_stored_summary_impl(summary.additional_summaries.and_then(|s| s.twb))`

rust/gitxetcore/src/summaries/analysis.rs
```diff
@@ -111,14 +114,18 @@ pub struct FileSummary {
     // for historical reasons this is called libmagic but does not use libmagic
     pub libmagic: Option<LibmagicSummary>,
 
+    // A buffer to allow us to add more to the serialized options
+    pub additional_summaries: Option<SummaryExt>,
+}
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
+pub struct SummaryExt {
+    pub version: u32,
```
hoytak, line 123: If the default version is 1 (given by ADDITIONAL_SUMMARY_VERSION above) then deriving Default will set this to 0.  Might just set ADDITIONAL_SUMMARY_VERSION to 0 or not derive Default?

hoytak (approved):
LGTM, I think this should do it.   Some coding conventions but the logic seems to match.

rust/gitxetcore/src/command/summary.rs
jgodlew: Not really, `summary` is a reference, so we need to have a reference `.as_ref()` to call the `and_then` (since `and_then` takes ownership of the option). If we grab the references for `additional_summaries` and `s.twb`, then we are stuck with an `Option<&TwbSummary>` instead of an `&Option<TwbSummary>` (&TwbSummary doesn't implement Default, only `TwbSummary`).

PR #310, opened 2024-03-26
Add backtrace to ErrorPrinter

seanses:
This makes the logging functions in ErrorPrinter to also include a backtrace.


hoytak (approved):
Looks okay to me, though in many places this will log it too far up the stack to be useful, but it shouldn't hurt.

PR #315, opened 2024-04-02
smudge and summarize

assafvayner:
Adds functionality to smudge from CAS hash then pass through to CSV/TWB/TDS file summarizers. With custom arg to specify delimiter since we don't have the file extension.


jgodlew (approved):
LGTM! 🚢

znation (approved):
LGTM.

Is the goal with this change to be able to summarize files that aren't in the repo? Otherwise, why not use the summarization from BlobID and point it to a pointer file blob? (I believe this should work by reading the pointer file and then computing summary from the smudged contents.)

PR #316, opened 2024-04-03
summarize compute from stdin and add force option for from-blob-id

assafvayner:
- Adds a git-xet summary subcommand to summarize file contents from stdin
- minor refactors of summary.rs functions, shared code between different compute command impls
- adds a `-f/--force` on `git-xet summary compute-from-blob-id` that skips attempting to look in summary DB in git notes
  - for blob that holds a pointer file, this will attempt to compute the hash by smudging the hash.


jgodlew (approved):
LLGTM! :shipit:

PR #87, opened 2023-08-01
Precheck and remove XET_NO_SMUDGE

seanses:
`XET_NO_SMUDGE` pollutes the config parsing from Env. This checks for env var `XET_NO_SMUDGE` and tries to remove it before parsing. See more context [here](https://github.com/xetdata/xethub/issues/3645#issuecomment-1659400324).

Fix https://github.com/xetdata/xethub/issues/3645.

PR #317, opened 2024-04-03
Handles no smudge config correctly

seanses:
Fixes CLI-144 [->](https://linear.app/xethub/issue/CLI-144/xet-no-smudge-doesnt-work-with-git-xet-clone). 

Context:
When cloning a repo "no smudge" can be configured from three places in order of descending priority:
1. command line args: `git xet clone --lazy/--no-smudge`
2. env var `XET_NO_SMUDGE=1`
3. git-xet global config, e.g. `git xet config --global smudge false`.

The bug was that `XET_NO_SMUDGE` was dropped from the env var but the config not used by `git xet clone`. It was dropped in https://github.com/xetdata/xet-core/pull/87 to avoid a config parsing issue which was later fixed by https://github.com/xetdata/xet-core/pull/95, making it unnecessary to drop `XET_NO_SMUDGE`.

This PR fixes the XetConfig to make sure it picks up the corresponding global config and env var config, and makes sure "git xet clone" uses that.

Tests:
1. `XET_NO_SMUDGE=1 git clone`: no smudge
2. `XET_NO_SMUDGE=1 git xet clone`: no smudge
3. `git xet config --global smudge false && git xet clone`: no smudge

### DIFF
--- a/rust/gitxetcore/src/command/clone.rs
+++ b/rust/gitxetcore/src/command/clone.rs
@@ -32,7 +32,8 @@ pub async fn clone_command(config: XetConfig, args: &CloneArgs) -> Result<()> {
     clone_xet_repo(
         Some(&config),
         &arg_v[..],
-        args.no_smudge || args.lazy,
+        // config.force_no_smudge may come from env var `XET_NO_SMUDGE` or git-xet global config `smudge`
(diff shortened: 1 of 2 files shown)


hoytak (approved):
LGTM

Issue #308, opened 2024-03-21
Please update the Cargo.lock file before releases


alerque:
I've been packaging releases for Arch Linux, and one thing we always try to do is create *reproducible* packages. That way anybody can verify that the binary artifacts in a release were indeed generated from the sources they were supposed to be. This also catches a lot of other problems with the build process.

One blocker for this project is that the Cargo.lock file is not getting updated properly. Ideally this this should be updated regularly during the development process, especially when any dependencies change in the Cargo.toml manifest or you want to pin new minor releases of dependencies. But **at the very least** it should be updated prior to tagging releases. That way the app can be built using `cargo build --locked` using a known-good set of dependency versions. This serves end users in case some dependency updates are inadvertently breaking, secures things in case of crate compromises, enables reproducible builds, etc.

Getting this done is pretty easy, just `cargo update` and commit the resulting `Cargo.lock` file. This should be done at a minimum after bumping the release version in `Cargo.toml`. It would probably be good to do it sometime shortly before the release, test with the pinned dependencies, then bump the version, then make sure the build runs with `cargo build --locked`.

CI should probably be adapted to run `--locked` too.

ylow:
Currently we are not checking in the Cargo.lock files as per Rust guidance. Though it looks like the policy has changed lately. https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html 
I still don't have a clear understanding of the implications and what this means, will ask around and get back to you in a bit.

alerque:
I think you've completely misunderstood the guidance, and it has not changed relative to this project. This project is not just a library, it is a whale app and ships several CLI binaries. The guidance on this scenario has *always* been to check in the lockfile. The only change is that the new guidance is that some pure-library crates should consider tracking lock files as well.

hoytak:
In our release environment, it's a little more complicated as we have an internal build process that mirrors some of what you see here. This is due to the need for signing on the windows and OSX releases and building matching server-side libraries that pair with the client release components.  We want one Cargo.lock for this larger piece.  

To work with your request, however, we will separate out the git-xet program from the general xet-core workspace, and this part will have its own Cargo.lock file checked in to the environment.  Hopefully this will make the build process easier on your end.

alerque:
[...]

PR #320, opened 2024-04-09
Move gitxet binary to separate folder out of workspace.

hoytak:
Also:
- Commit Cargo.lock to gitxet/ package.
- Add libxet shim library so we can slowly move dependencies on top of that.


seanses (approved):

hoytak:
Addresses https://github.com/xetdata/xet-core/issues/308.

alerque:
I don't think this is necessary. See [this comment](https://github.com/xetdata/xet-core/issues/308#issuecomment-2048212977).

PR #321, opened 2024-04-10
Adding simple summary diffs for twb and tds files

jgodlew:
Fixes: TAB-47

* Adds a summary diff processor for twb files that just provides the before/after summaries
* Adds a summary diff processor for tds files that just provides the before/after summaries

### DIFF
--- a/rust/gitxetcore/src/diff/twb.rs
+++ b/rust/gitxetcore/src/diff/twb.rs
@@ -1,0 +1,63 @@
+use serde::{Deserialize, Serialize};
+use tableau_summary::twb::TwbSummary;
+use crate::diff::output::SummaryDiffData;
+use crate::diff::{DiffError, SummaryDiffProcessor};
+use crate::diff::output::SummaryDiffData::Twb;
+use crate::summaries::{FileSummary, SummaryType};
+
+/// Diff content for a Twb diff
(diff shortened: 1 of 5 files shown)


rust/gitxetcore/src/diff/twb.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    fn get_remove_diff(&self, summary: &TwbSummary) -> Result<SummaryDiffData, DiffError> {
+        Ok(Twb(TwbSummaryDiffContent {
+            before: Some(summary.clone()),
+            after: None,
+        }))
+    }
+
+    fn get_diff_impl(
```
sirahd, line 53: [q] It looks like TWB/TDS diff processor simply copies the before and after summary into its struct. Do we expect the diffing logic (i.e. figuring out which worksheets/columns/types are added, changed or removed) to be implemented in Xetea?

sirahd (approved):

rust/gitxetcore/src/diff/twb.rs
jgodlew: Yeah, ideally, we will have the actual diff logic implemented in here. For now, we are just displaying the before/after summaries on the UI. We could have Xetea perform the diff given this before/after struct (which is what CSV summary diffs do), but that probably won't work for us since the summary structure is so complex.

PR #322, opened 2024-04-10
match cas/shard client hyper versions

assafvayner:
bump both cas_client & shard_client packages to have dependencies hyper v1.2.0 and http v1.1.0


hoytak (approved):
LGTM!  Sorry I missed this.

PR #325, opened 2024-04-18
Fixing possible unaligned exception on client.

hoytak:
Uses operations that will be correct universally. 

Fixes @CLI-166.


seanses (approved):
LGTM!

PR #328, opened 2024-04-19
Remove automatic git path discovery when not in git repository.

hoytak:
Removes paths that automatically discover git remotes and config information from the current directory when config is not provided.  

Now, the repo path is set in XetConfig at the start, possibly from the current directory, but parts of the code that previously just defaulted to automatic discovery have been reworked.

### DIFF
--- a/rust/gitxetcore/src/xetblob/xet_repo.rs
+++ b/rust/gitxetcore/src/xetblob/xet_repo.rs
@@ -224,7 +224,7 @@ impl XetRepo {
         let remote = if let Some(repo_info) = repo_info.as_ref() {
             repo_info.repo.html_url.clone()
         } else {
-            let remotes = config.remote_repo_paths();
+            let remotes = config.known_remote_repo_paths();
             if remotes.is_empty() {
                 return Err(anyhow!("No remote defined"));
             }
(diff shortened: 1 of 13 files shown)


rust/gitxetcore/src/xetblob/xet_repo.rs
```diff
@@ -224,7 +224,7 @@ impl XetRepo {
         let remote = if let Some(repo_info) = repo_info.as_ref() {
             repo_info.repo.html_url.clone()
```
seanses, line 225: I think we need to pass this url into config so that `gitxetcore::data::cas_interface::create_cas_client` can pick it up and use it for billing purpose

seanses (approved):
We'll get @junpengxxx here to confirm where billing is getting repo paths.

rust/gitxetcore/src/xetblob/xet_repo.rs
hoytak: With discussion offline -- it's likely this is the case here, although user and auth information is populated in the config struct.   We need to clean that up for sure, but we'll get this PR in here and do that in the next one.

PR #329, opened 2024-04-19
Hide git checkout command.

hoytak:
Hides the slightly broken, poorly documented, and not really used `git xet checkout` command


seanses (commented):
Can we also hide Filter, Push, Hooks, Summary, DirSummary, Diff, VisualizationDependencies?

seanses (approved):

PR #331, opened 2024-04-20
Finer-grained diff for TWB files.

jgodlew:
Fixes: TAB-71, TAB-72, TAB-73

* Moves twb diff calculation to tableau_summary crate
* New diff calculation to perform finer-grained diffs on the TwbSummary structs to provide a richer diff.
* Utility trait (`DiffProducer`) to help produce diffs from a Summary type
* Generic `DiffItem` for holding the before/after/status info for some type (e.g. a string).
* Diffs implemented for twb datasources, worksheets, and dashboards.

TODO:
- [x] There are a lot of layers/enums. Maybe we can reduce this.


rust/tableau_summary/src/twb/diff/dashboard.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub num_changes: usize,
+    pub name: DiffItem<String>,
+    pub title: DiffItem<String>,
+    pub thumbnail: DiffItem<Option<String>>,
+    pub sheets: Vec<DiffItem<String>>,
+    pub zones: ZoneDiff,
+}
+
+impl DashboardDiff {
+    fn update_num_changes(&mut self) {
```
port8080, line 17:
```suggestion
    fn calculate_num_changes(&mut self) {
```
It looks to me that this is always called when the changes are zero and then they are calculated. If that is the case, the `update` in the name is a bit confusing

port8080 (approved):

PR #335, opened 2024-04-25
Added streaming process to form a data iterator from stdout.

hoytak:
Adds an AsyncDataIterator implementation that streams data chunks from the stdout of an external process.  This allows the stdout of a `git lfs smudge` or `git xet smudge` command to be streamed directly into our data ingestion commands.

### DIFF
--- a/rust/gitxetcore/src/stream/stdout_process_stream.rs
+++ b/rust/gitxetcore/src/stream/stdout_process_stream.rs
@@ -46,7 +46,7 @@ impl AsyncStdoutDataIterator {
             unreachable!();
         };
 
-        stdin.write_all(&stdin_data).await?; // Send all data
+        stdin.write_all(stdin_data).await?; // Send all data
         stdin.flush().await?; // Ensure all data is sent
 
         Self::from_process(process, stdout_bufsize)
--- a/rust/gitxetcore/src/stream/stdout_process_stream.rs
+++ b/rust/gitxetcore/src/stream/stdout_process_stream.rs
@@ -11,8 +11,6 @@ use parutils::AsyncIterator;
 
 use super::data_iterators::AsyncDataIterator;
 
-const CHILD_PROCESS_POLL_INTERVAL_MILLIS: u64 = 250;
-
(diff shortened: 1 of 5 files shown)


rust/gitxetcore/src/stream/stdout_process_stream.rs
```diff
@@ -0,0 +1,285 @@
...
+        if readlen > 0 {
+            buffer.resize(readlen, 0);
+            Ok(Some(buffer))
+        } else {
+            let Some(child) = self.child_process.take() else {
+                Err(anyhow!("Next called after child process completed."))?;
+                unreachable!();
+            };
+
+            let output = child.wait_with_output().await?;
```
seanses, line 94: I smell a bit danger here. `read` returning 0 does not mean that the reader will always no longer be able to produce bytes so there may be remaining bytes in this output.

hoytak: Good catch.  Updated this and read up on all the cases where we should do a retry.  Rust's io read api is dumb.

ylow (approved):

rust/gitxetcore/src/stream/stdout_process_stream.rs
```diff
@@ -0,0 +1,403 @@
...
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        // First wait for the child process
+        let Some(child) = self.child_process.take() else {
+            Err(anyhow!("Next called after child process completed."))?;
```
seanses, line 120: Maybe return EOF error instead of anyhow because this is still legit termination, so callers can check for this error.

seanses (approved):

rust/gitxetcore/src/stream/stdout_process_stream.rs
hoytak: In this case, it would be a programming error -- EOF should always turn into None in the iterator, so they shouldn't be checking for EOF for that case.   This one only happens when the iterator gets reused after it sends a termination.

PR #336, opened 2024-04-25
Sketches for on versioning.

hoytak:


hoytak:
Was example, not to be merged.

PR #337, opened 2024-04-25
git xet migrate

hoytak:
(BETA)  Migrate an external repository to a new XetHub repository.   All commits, branches, and other files are converted, history is fully preserved, and all data files stored as LFS or Xet pointer files are imported.

```
USAGE:
    git-xet migrate [OPTIONS] --src <SRC> --dest <DEST>

OPTIONS:
        --dest <DEST>                  The URL of the Xet repository to initialize with the contents
                                       of the src repository. This repository must be a new Xet
                                       repo, or use --overwrite to replace all contents of this
                                       repository with the source repo
    -h, --help                         Print help information
        --no-cleanup                   Do not clean up the working directories after completion
        --overwrite                    The migrate command will error out if the remote repository
                                       is not empty. When this flag is passed, the contents of the
                                       remote repository will be overwritten regardless of state
        --src <SRC>                    The URL of the git repository to import
    -V, --version                      Print version information
        --working-dir <WORKING_DIR>    The directory to use to do all of the processing in (default:
                                       ~/.xet/migration)
                                       
```

### DIFF
--- a/rust/gitxetcore/src/command/migrate.rs
+++ b/rust/gitxetcore/src/command/migrate.rs
@@ -2,7 +2,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use clap::Args;
-use tracing::error;
+use tracing::{error, warn};
 
 use crate::config::XetConfig;
 use crate::errors::Result;
@@ -80,18 +80,47 @@ pub async fn migrate_command(config: XetConfig, args: &MigrateArgs) -> Result<()
     )?;
 
     if !args.overwrite {
-        // Check to make sure it's a new repository.
-        let (_, commit_count, _) = run_git_captured(
+        // Check to make sure it's a new repository.  This is using xethub's default repo as a model, so it's not
+        // the most robust.  On any error, just ask the user to use --overwrite
+        let commit_count = run_git_captured(
             Some(&dest_dir),
             "rev-list",
-            &["--count", "HEAD"],
+            &["--count", "main"],
             true,
             None,
-        )?;
+        )
+        .map_err(|e| {
+            warn!("Error getting commit count for testing if repo is new: {e:?}");
+            e
+        })
+        .ok()
+        .and_then(|(_, commit_count_s, _)| commit_count_s.parse::<usize>().ok())
+        .unwrap_or(usize::MAX);
 
-        if commit_count.parse::<usize>().unwrap_or(1) > 1 {
+        if commit_count > 1 {
             return Err(crate::errors::GitXetRepoError::InvalidOperation(format!(
-                "ERROR: Xet Repository {} not a new repository; refusing to overwrite changes.",
+                "ERROR: Xet Repository {} has multiple commits on main; refusing to overwrite changes.  Use --overwrite to force operation.",
+                &args.dest
+            )));
+        }
+
+        let has_nonmain_branches = run_git_captured(
+            Some(&dest_dir),
+            "branch",
+            &["-l", "--format=%(refname:short)"],
+            true,
+            None,
+        )
+        .map_err(|e| {
+            warn!("Error getting branch list for repo: {e:?}");
+            e
+        })
+        .map(|(_, branches, _)| branches.lines().any(|c| c.trim() != "main"))
+        .unwrap_or(true);
+
+        if has_nonmain_branches {
+            return Err(crate::errors::GitXetRepoError::InvalidOperation(format!(
+                "ERROR: Xet Repository {} has multiple branches; refusing to overwrite.  Use --overwrite to force operation.",
                 &args.dest
             )));
         }
--- a/rust/gitxetcore/src/command/migrate.rs
+++ b/rust/gitxetcore/src/command/migrate.rs
@@ -31,10 +31,6 @@ pub struct MigrateArgs {
     #[clap(long)]
     pub no_cleanup: bool,
 
-    /// Skip checking out the repository at the end.
-    #[clap(long)]
-    pub skip_check: bool,
-
     /// The directory to use to do all of the processing in (default: ~/.xet/migration).
     #[clap(long)]
     pub working_dir: Option<String>,
@@ -51,6 +47,8 @@ pub async fn migrate_command(config: XetConfig, args: &MigrateArgs) -> Result<()
     };
 
     std::fs::create_dir_all(&working_dir)?;
+
+    let working_dir = working_dir.canonicalize()?;
 
     let source_dir = working_dir.join("source");
     let dest_dir = working_dir.join("xet_repo");
@@ -129,8 +127,9 @@ pub async fn migrate_command(config: XetConfig, args: &MigrateArgs) -> Result<()
         true,
         Some(&[("XET_DISABLE_HOOKS", "0")]),
     )?;
+
     // This command may fail due to the --all (504 error) as sometimes it overwhelms the endpoint?  So
-    // run regular push first, then push all the branches.
+    // run regular push first, then push all the branches.  This seems to work consistently
     run_git_passthrough(
         Some(&dest_dir),
         "push",
--- a/rust/gitxetcore/src/command/migrate.rs
+++ b/rust/gitxetcore/src/command/migrate.rs
@@ -46,7 +46,7 @@ pub async fn migrate_command(config: XetConfig, args: &MigrateArgs) -> Result<()
         }
     };
 
-    std::fs::create_dir_all(&working_dir)?;
+    config.permission.create_dir_all(&working_dir)?;
 
     let working_dir = working_dir.canonicalize()?;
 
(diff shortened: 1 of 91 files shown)


rust/gitxetcore/src/data/pointer_file.rs
```diff
@@ -178,6 +178,18 @@ impl PointerFile {
     }
 }
 
+pub fn is_xet_pointer_file(data: &[u8]) -> bool {
+    if data.len() >= 128 {
```
seanses, line 182: Recommend not to use "128" although we know xet pointer file size is never going to exceed 128. Maybe switch to "POINTER_FILE_LIMIT" or any other constant var with comments.

rust/gitxetcore/src/command/migrate.rs
```diff
@@ -0,0 +1,158 @@
...
+    let working_dir = {
+        if let Some(wd) = args.working_dir.as_ref() {
+            PathBuf::from_str(wd).unwrap()
+        } else {
+            let migration_tag = uuid::Uuid::now_v6(&[0; 6]).to_string();
+            config.xet_home.join("migration").join(migration_tag)
+        }
+    };
+
+    std::fs::create_dir_all(&working_dir)?;
```
seanses, line 49: For dirs and files under xet_home, recommend to create using `config.permission.create_*` so there's no risk of root ownership when running the command under `sudo`.

rust/gitxetcore/src/command/migrate.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let source_dir = working_dir.join("source");
+    let dest_dir = working_dir.join("xet_repo");
+
+    eprintln!("XET: Retrieving Source Repo {:?}", &args.src);
+
+    run_git_passthrough(
+        Some(&working_dir),
+        "clone",
+        &["--mirror", &args.src, source_dir.to_str().unwrap()],
```
seanses, line 61: Do they meant to be cloned with different option `--mirror` for src vs `--bare` for dest? `--mirror` sets up something more than `--bare` but I don't remember what exactly.  Are we also migrating those?

rust/gitxetcore/src/command/migrate.rs
```diff
@@ -0,0 +1,158 @@
...
+        // Check to make sure it's a new repository.
+        let (_, commit_count, _) = run_git_captured(
+            Some(&dest_dir),
+            "rev-list",
+            &["--count", "HEAD"],
+            true,
+            None,
+        )?;
+
+        if commit_count.parse::<usize>().unwrap_or(1) > 1 {
```
seanses, line 89: What if `HEAD` is empty (only one commit) but other branches are not

rust/gitxetcore/src/command/migrate.rs
```diff
@@ -0,0 +1,158 @@
...
+            return Err(crate::errors::GitXetRepoError::InvalidOperation(format!(
+                "ERROR: Xet Repository {} not a new repository; refusing to overwrite changes.",
+                &args.dest
+            )));
+        }
+    }
+
+    // Now, push everything to the remote.
```
seanses, line 97: This comment is not consistent with the next segment of code

rust/gitxetcore/src/command/migrate.rs
hoytak: Good catch, will do!

rust/gitxetcore/src/command/migrate.rs
hoytak: Yes, this distinction is intentional.  --mirror copies all the references as they are, and is faster than --bare.  Since we're just matching the remote, this is totally fine for us here.    --bare in the dest clone processes all the references and sets up new ones locally so that we can do the force push at the end.

rust/gitxetcore/src/command/migrate.rs
hoytak: This is currently assuming that the repo is set up exactly how xethub would, so I'm not really considering other cases there.  If this isn't the case, then we can revisit that assumption.

rust/gitxetcore/src/command/migrate.rs
hoytak: Added comments.

rust/gitxetcore/src/command/migrate.rs
hoytak: Logic updated now tests that main is the only branch and has exactly 1 commit.

rust/gitxetcore/src/git_integration/repo_migration.rs
```diff
@@ -0,0 +1,675 @@
...
+            tr_map.insert(b_oid, new_id);
+            progress_reporting.register_progress(Some(1), None);
+        }
+
+        pft.finalize().await?;
+        progress_reporting.finalize();
+    }
+
+    // Create a gitattributes blob object, if it's not there already.
+    let gitattributes_oid = dest.blob(GITATTRIBUTES_CONTENT.as_bytes())?;
```
seanses, line 176: Does this function return error if it's already there?

rust/gitxetcore/src/git_integration/repo_migration.rs
```diff
@@ -0,0 +1,675 @@
...
+                        }
+                    }
+
+                    // Add the translated Oid to the new Tree with the same name and filemode
+                    tree_builder.insert(entry.name_bytes(), *new_oid, entry.filemode_raw())?;
+                }
+
+                if is_root {
+                    // Add in the .gitattributes entry explicitly, as this is a root commit.
+                    tree_builder.insert(".gitattributes", gitattributes_oid, 0o100644)?;
```
seanses, line 301: This implies that checking out commits that are not at the head of references will not get xet pointer files automatically materialized. I'm OK with that if the product team agrees, just need to clearly state that in docs.

rust/gitxetcore/src/git_integration/repo_migration.rs
hoytak: No, if it's already present there isn't an issue.

rust/gitxetcore/src/git_integration/repo_migration.rs
```diff
@@ -0,0 +1,675 @@
...
+
+                let Some(new_tree_id) = tr_map.get(&src_commit.tree_id()) else {
+                    panic!("Logic Error: commit tree id not translated.");
+                };
+
+                let new_tree = dest.find_tree(*new_tree_id)?;
+
+                // Create a new commit in the destination repository
+                let new_commit_id = dest.commit(
+                    None, // This updates the HEAD to point to the new commit
```
seanses, line 395: I noticed that API instead says `Use "HEAD" to update the HEAD of the current branch and make it point to this commit`, change the comment if this is desired

seanses: Feature request: is it possible to keep the original commit timestamp? I would be very happy to see my history actually doesn't change. But this may be a product question

rust/gitxetcore/src/git_integration/repo_migration.rs
hoytak: I don't think that's the logic here, as we convert all commits in the repo by default.  root here means a tree in the root directory of the repo, meaning the tree that represents the base directory.   We ensure that for all commits in the repo, <repo>/.gitattributes is set.

As a side note, orphaned commits are not really accessible anyway except by doing some really deep git searching unless you know them already.   By default, `git gc` deletes such commits if they are more than 2 weeks old.

rust/gitxetcore/src/git_integration/repo_migration.rs
hoytak: The comment here is wrong -- will update.  None does not update HEAD. 

It does, however, make me want to ensure that HEAD is indeed set to main.

seanses (commented):
This is great work! Left some comments above. 

One request: this translation holds a tree DAG, a commit DAG, a oid map and some auxiliary structures to do a topological sort, what's the RAM complexity? It will be great to convey this to users and also to fail early on large repo migrations that exceed RAM limit.

rust/gitxetcore/src/git_integration/repo_migration.rs
hoytak: The timestamp is contained in the committer signature object, which we just copy over, so it should be preserved.

hoytak:
Currently, only the OID (20 bytes) and indexing are stored in memory, so I don't think that would be an issue.   It was something I thought about though. 

If we assume that there is another 80 bytes per OID between references, hash table overhead, etc., (generous) then we would be able to store 10,000,000 objects per GB of ram, which is already an enormous repo, so I don't think this should be an issue.   
Also, the repo itself would be much larger, as that doesn't include any of the content.

rust/gitxetcore/src/git_integration/repo_migration.rs
seanses: Ah, yes you're right! Got confounded by "root"

seanses (approved):
LGTM!

PR #338, opened 2024-04-25
Update gitxet/Cargo.toml version string

hoytak:
The executable in gitxet/ did not get the version string properly updated.  This PR fixes this.


seanses (commented):
And libxet?

We should also make the changes to the CM template

hoytak:
> And libxet?

Good catch! 
> 
> We should also make the changes to the CM template

Already did that. (Though I think we should add a script here too.)

seanses (approved):

PR #339, opened 2024-04-25
Move pyxet into xet-core

hoytak:
Pyxet development is connected to xet-core development, and having a separate python repository has introduced unneeded friction in the development process.  In addition, some backend errors were introduced recently that were only caught in the pyxet unit tests, so running these as part of the regular xet-core development will make the build process more robust.   

This PR takes a source drop of pyxet, switches the dependency path of pyxet to be the local directory (through libxet), and merges the github actions of the two repositories.


.github/workflows/commit.yml
```diff
@@ -140,16 +140,15 @@ jobs:
         if: runner.os != 'macOS'
         run: |
           cd python/pyxet
-          maturin build -r
+          maturin build --debug
```
seanses, line 143: I think no option specified means debug build. cargo apparently complains about "error: unexpected argument '--debug' found"

.github/workflows/commit.yml
```diff
@@ -140,16 +140,15 @@ jobs:
...
       - name: Build with maturin (macOS)
         if: runner.os == 'macOS'
         run: |
           brew install make
           export PATH="/usr/local/opt/make/libexec/gnubin:$PATH"
           export CXXFLAGS="-stdlib=libc++"
           cd python/pyxet
-          rustup target add aarch64-apple-darwin x86_64-apple-darwin 
-          maturin build --target universal2-apple-darwin -r --features openssl_vendored 
+          maturin build --debug --features openssl_vendored
```
seanses, line 151: Same as above.

seanses (approved):
I'm 👍 with this, given that we are going to install git-xet and xet cli together we should also release them in lockstep. May need to confirm with product team over branding and marketing questions.

seanses:
We'll also need to copy over the action secrets and env variables to run the pyxet tests.

hoytak:
> We'll also need to copy over the action secrets and env variables to run the pyxet tests.

I believe these variables are account wide, so we should be fine.

.github/workflows/commit.yml
hoytak: Yes, typo!

hoytak:
Vetoed by @ylow, closing.

PR #343, opened 2024-04-30
Tableau Summary Refactoring

jgodlew:
Fixes: TAB-83

Refactors the TwbSummary and TdsSummary to use a versioned enum for the summary. This will allow us to add new schemas for the summaries as needed while remaining backwards compatible with older summaries.

Details of testing can be found in: https://www.notion.so/xethub/Summary-Versioning-Compatibility-21f173305a9c44dea05c33ed5b176546. 

Note that old clients won't be able to parse newer schemas (e.g. V2+).

TODO:
- [x] Update TdsSummary to follow the same enum versioning strategy.


rust/gitxetcore/src/summaries/summaries_plumb.rs
```diff
@@ -316,3 +316,196 @@ pub async fn summaries_dump(config: XetConfig) -> errors::Result<()> {
     println!("{json}");
     Ok(())
 }
+
+#[cfg(test)]
```
hoytak, line 320: I'm not sure that this test format is a good idea here...  The issue is that there is no guaranteed ordering around which tests run first, especially when they are run as part of the larger process.   Also, it would be better to write out temp files to a tempfile.TempDir() path, which gets cleaned up when the object is destroyed.  

Looking at these tests, however, do you need a temporary file for most of them?  It seems like you could have the serialization function just return a Vec<u8> and then pass that to the deserialization routines, correct?

rust/gitxetcore/src/diff/csv.rs
```diff
@@ -51,8 +52,8 @@ impl SummaryDiffProcessor for CsvSummaryDiffProcessor {
         1
     }
 
-    fn get_data<'a>(&'a self, summary: &'a FileSummary) -> Option<&Self::SummaryData> {
-        summary.csv.as_ref()
+    fn get_data<'a>(&'a self, summary: &'a FileSummary) -> Option<Cow<Self::SummaryData>> {
```
hoytak, line 55: It feels like the Copy on Write mechanic here is overkill -- what's the need to use that rather than just a reference and then an explicit clone() if a writeable version is needed?

rust/gitxetcore/src/summaries/summaries_plumb.rs
jgodlew: So the issue is that I needed to have the serialization and deserialization happen on different versions of the codebase (i.e. main, this-branch, v2-branch) since each branch produces a different DB and I want to evaluate backwards and forwards compatibility. I can't run these tests automatically since I can't change the branch of this repo dynamically inside the test (thus the serialization to a permanent file). 
Note that all of the tests are `#[ignored]` so they don't run automatically.

However, I might just remove these tests since I've saved the testing code and results in an investigation doc: https://www.notion.so/xethub/Summary-Versioning-Compatibility-21f173305a9c44dea05c33ed5b176546

rust/gitxetcore/src/diff/csv.rs
hoytak: Ah, I see...  I kinda wonder if it would just be better to use Arc<...> inside the struct, with get_data() returning Arc<..>?  I feel like the ownership model of Cow is a bit weird here, as once created we don't really intend to write it (unless I'm wrong there).

rust/gitxetcore/src/summaries/summaries_plumb.rs
hoytak: Chatted offline about this and have a solid way forward for testing down the way.  Getting this in now.

rust/gitxetcore/src/diff/csv.rs
hoytak: Chatted offline about this; going with Cow for now.  A better model may be available in the future.

hoytak (approved):

PR #345, opened 2024-05-02
Fix Zone change count 

jgodlew:
Fixes: TAB-86

Changes the count of zone changes to only be 1 if any of the fields changes.


bpronan (approved):
Looks good. Thanks for fixing it!

PR #347, opened 2024-05-06
Changing the tableau diff summary schema to v1

bpronan:


bpronan:
This change is going into https://github.com/xetdata/xet-core/pull/348.

PR #348, opened 2024-05-06
Add table relations to Tableau Datasource summaries.

jgodlew:
Fixes: TAB-79, TAB-75

- Adds table relationships to the Datasource summary (logical table relationships).
- This produces a new V2 schemas for Twb + Tds summaries.
- Adds diffs of relationships to diff schema.

### DIFF
--- a/rust/tableau_summary/src/twb/summary/datasource.rs
+++ b/rust/tableau_summary/src/twb/summary/datasource.rs
@@ -1,8 +1,17 @@
 use std::collections::HashMap;
+use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use tracing::{error, info};
 use crate::twb::raw::datasource::{RawDatasource, substituter};
 use crate::twb::summary::util;
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
+pub struct DatasourceV1 {
+    pub name: String,
+    pub version: String,
+    pub tables: Vec<Table>,
+    pub added_columns: Option<Table>,
+}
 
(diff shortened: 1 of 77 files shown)


rust/tableau_summary/src/twb/summary/datasource.rs
```diff
@@ -1,15 +1,30 @@
...
+#[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
+pub struct DatasourceV1 {
+    pub name: String,
+    pub version: String,
+    pub tables: Vec<Table>,
+    pub added_columns: Option<Table>,
+}
+
 #[derive(Serialize, Deserialize, Default, PartialEq, Clone, Debug)]
 pub struct Datasource {
```
sirahd, line 22: Does it make sense to call this `DatasourceV2` to avoid having to rename `Datasource` -> `DatasourceV2` with the expectation that it may be updated in the future?

sirahd (approved):
LGTM!

bpronan (approved):
Looks good to me.

rust/tableau_summary/src/twb/summary/datasource.rs
jgodlew: I decided to make the "latest" schema unversioned (i.e. `Datasource`) to reduce the changes needed when updating. 

If I changed to `DatasourceV2`, it starts to become unclear which structs should have a `V*` appended to them: should I version every Summary struct in the event that we decide we might need to change it in the future (e.g. should I have `TableV1` instead of `Table` or a `DashboardV1`)? I would rather keep the convention of un-versioned meaning latest and a version suffix being an older version of the struct.

PR #349, opened 2024-05-06
Upgrade blake3 for optimized Arm64 / M1 support.

hoytak:
As of version 1.4, blake3 now has optimized M1 / Neon support.  This PR upgrades our version to take advantage of those speedups.


seanses (approved):
🚀

PR #350, opened 2024-05-06
Adding tds diff v1

jgodlew:
Fixes: TAB-74

Changes the diffs for Tds files to provide more detailed information (i.e. the same as twb datasource diffs).

(Note: will need to apply changes to xdiff once merged).


bpronan (approved):
Looks good to me.

This will require a corresponding front end change for TDS diffs to work. It should be easy to do, but maybe we wait on that being done before we merge this.

PR #351, opened 2024-05-07
Speed up windows PR tests

hoytak:
Currently, the windows PR tests are really slow due to a LOT of bash shell work to generate parts of the test stuff.  This PR adds a couple of test utility programs in rust,  xettest_create_file and xettest_create_csv, that are much faster on the windows bash emulation layer.


hoytak:
Nope, not actually helping.  Seems they are just slow regardless.

PR #353, opened 2024-05-07
Moved `git-xet migrate` to `git xet repo migrate`

hoytak:
In discussion with @erinys, change API of `git xet migrate` to `git xet repo migrate` in order to allow for more repo level commands in the future.  API is the same for the command itself.


seanses (approved):
LGTM!

PR #355, opened 2024-05-08
Fix compression reporting bug

seanses:
lz4 compressed size can be larger than the original size. For example, I added `10/` and `11/` under https://xethub.com/XetHub/Flickr30k/src/branch/main/flickr30k_images and got:
```
di@di-mbp ~/tt/int64 % git add .
Git-Xet 0.13.4: Processing data: 76.37 MiB | 23.10 MiB/s, done.
76.37 MiB added, stored 76.63 MiB (-0.3% reduction)
```


sirahd (approved):

hoytak (approved):
Fine by me.

PR #357, opened 2024-05-08
Overwrite xetblob file on parse failure

seanses:
We added an extra field `cas` to the RepoInfo struct. `serde` parsing old saved info will fail. Tested locally that this will overwrite the repo metadata file on parse failure and then carry on operations correctly.

### DIFF
--- a/rust/gitxetcore/src/xetblob/xet_repo.rs
+++ b/rust/gitxetcore/src/xetblob/xet_repo.rs
@@ -283,13 +283,12 @@ impl XetRepo {
     ) -> anyhow::Result<()> {
         let xet_repo_info_file = path.join(XET_REPO_INFO_FILE);
 
-        let saved_repo_info: Option<RepoInfo> = if xet_repo_info_file.exists() {
-            Some(serde_json::de::from_slice(&std::fs::read(
-                &xet_repo_info_file,
-            )?)?)
-        } else {
-            None
-        };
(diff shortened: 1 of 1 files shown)


rust/gitxetcore/src/xetblob/xet_repo.rs
```diff
@@ -280,16 +280,16 @@ impl XetRepo {
...
-                &xet_repo_info_file,
-            )?)?)
-        } else {
-            None
-        };
+        // Try to parse a RepoInfo struct from maybe some bytes from a maybe existing metadata file.
+        // Any failure leads to overwriting the file.
+        let saved_repo_info = std::fs::read(&xet_repo_info_file)
+            .map(|bytes| serde_json::de::from_slice::<RepoInfo>(&bytes).ok())
+            .ok()
```
hoytak, line 291: Can use and_then() instead of map to tighten this up.

rust/gitxetcore/src/xetblob/xet_repo.rs
```diff
@@ -303,7 +303,7 @@ impl XetRepo {
         // repo changed, delete all contents and write out the repo info
         if repo_changed {
             std::fs::remove_dir_all(path)?;
-            std::fs::create_dir_all(path)?;
+            config.permission.create_dir_all(path)?;
```
hoytak, line 306: Nice

hoytak (approved):
Fix up details + lint, but lgtm

PR #358, opened 2024-05-09
Better list diffs for Twb/Tds summaries

jgodlew:
Fixes: TAB-87

Switches diffs to use the Myers diff algorithm implemented using the [imara-diff](https://github.com/pascalkuthe/imara-diff) crate, allowing for minimal diffs between two lists of items.

* Summary/Diff structs  need to have `Eq + Hash` traits defined.
* Adds a `DiffTokenSource` to provide tokens from a list of some T to the diff algorithm.
* Adds a `DiffSync` to process "hunks" of changes that the diff algorithm produces (e.g. `before[1..3]` and `after[2..4]` differ).

### DIFF
--- a/rust/tableau_summary/src/twb/diff/util.rs
+++ b/rust/tableau_summary/src/twb/diff/util.rs
@@ -192,7 +192,7 @@ pub trait DiffProducer<T>: Sized {
         where
             T: Eq + Hash
     {
-        let input = InternedInput::new(DiffTokenSource::new(before), DiffTokenSource::new(after));
+        let input = InternedInput::new(DiffTokenSource(before), DiffTokenSource(after));
         imara_diff::diff(MyersMinimal, &input, DiffSink::new(&input))
     }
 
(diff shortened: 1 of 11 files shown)


rust/tableau_summary/src/twb/diff/util.rs
```diff
@@ -341,21 +478,120 @@ mod tests {
             vec!["bar", "baz"],
             vec!["foo", "bar", "baz"],
             vec![
-                (ChangeState::Change, Some("bar"), Some("foo")),
-                (ChangeState::Change, Some("baz"), Some("bar")),
-                (ChangeState::Add, None, Some("baz")),
+                (ChangeState::Add, None, Some("foo")),
+                (ChangeState::None, Some("bar"), None),
+                (ChangeState::None, Some("baz"), None),
```
jgodlew, line 483: This is an example of the simplification this PR makes:
when prepending "foo" to the list, instead of that being treated as "2 changes and an add", it is just an "add and two no-ops".

sirahd (approved):
Very cool! I had to read up this blog post to learn more about Myers diff algorithm: https://blog.jcoglan.com/2017/02/12/the-myers-diff-algorithm-part-1/

PR #360, opened 2024-05-09
Special case public workbooks

jgodlew:
Fixes: TAB-116

Updates thumbnail-name to special-case public.tableau.com since the format of the `repository_location` tag uses the `id` column for the name instead of the last path element of the url.

### DIFF
--- a/rust/tableau_summary/src/twb/raw/util.rs
+++ b/rust/tableau_summary/src/twb/raw/util.rs
@@ -14,12 +14,26 @@ pub fn parse_formatted_text(n: Node) -> String {
 pub fn repository_location_to_thumbnail_name(n: Node) -> String {
     let id = n.get_attr("id");
     let derived_from = n.get_attr("derived-from");
-    let view_name = Url::parse(&derived_from)
-        .map(|u| u.path_segments()
-            .and_then(|s| s.last())
-            .map(String::from)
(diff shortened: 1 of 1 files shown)


rust/tableau_summary/src/twb/raw/util.rs
```diff
@@ -14,12 +14,26 @@ pub fn parse_formatted_text(n: Node) -> String {
...
+                .and_then(|s| s.last())
+                .map(String::from),
+            u.host_str()
+                .map(str::to_string)
+        )
+        ).unwrap_or((None, None));
+    if let Some(hostname) = hostname {
+        // the viewname for public.tableau.com urls is a number (id in public),
+        // we need the view name, which is in the id instead.
+        if hostname.contains("public.tableau.com") {
```
bpronan, line 29: I was about to recommend that we make this magic value a constant, but it would lose the context of the comment. 

I'll just highlight it and leave it up to you to make a constant if you prefer.

bpronan (approved):
Looks good!

rust/tableau_summary/src/twb/raw/util.rs
jgodlew: No problem, I think it is fine to make it a constant.

PR #362, opened 2024-05-10
Change zones to be allow multiple top-level Zone objects

jgodlew:
Fixes: TAB-115

Since the `<zones>` tag might have multiple `<zone>` tags within, we need to change the Raw and Summary Dashboards to support this.

Note: this bumps up the schema version of the TwbSummary to 3 since this is a type change. We should make sure that the UI is able to accept these V3 schemas before merging.


bpronan (approved):
Looks good and makes sense. 

I'll try to address the front-end changes this afternoon.

sirahd (approved):

PR #363, opened 2024-05-14
Fixes for git xet migrate 502 transport error.

hoytak:
This PR adds a fix for the 502 transport error for when it seems pushing all the branches gets 
```
error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
send-pack: unexpected disconnect while reading sideband packet
fatal: the remote end hung up unexpectedly
```
This fixes one possible cause, but then as backup pushes each branch individually.

### DIFF
--- a/gitxet/Cargo.lock
+++ b/gitxet/Cargo.lock
@@ -29,6 +29,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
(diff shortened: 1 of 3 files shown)


seanses (approved):
LGTM! The `http.postBuffer` [doc](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httppostBuffer) reads a bit alarming, but IIUC that's only a temp buffer size on the client side (not the max pack file size)?

hoytak:
> LGTM! The `http.postBuffer` [doc](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httppostBuffer) reads a bit alarming, but IIUC that's only a temp buffer size on the client side (not the max pack file size)?

Yes, that's correct, and only when preparing uploads.  Max pack file size is different.

PR #368, opened 2024-05-14
Fixes significant slowdown when chunking huge buffers of constant data.

hoytak:
When using large buffers in the low variance chunking algorithm, the chunker will scan all the way to the end of the buffers even when it's way past the max_chunk_size.  This can cause near-quadratic behavior.  This PR limits the read to the max that can be part of the next chunk.


ylow (approved):

ylow:
Took me a while to page in the context around it. Good catch!

PR #370, opened 2024-05-16
Limit clean task memory usage

seanses:
Fix CLI-190

Given the fact that 
1. Git and PyXet send data in different block size to xet-core to clean,
2. PyXet can clean multiple files in parallel,
3. Different files have different clean speed,

it's difficult to tune memory usage for each clean task. Instead, clean tasks can share a common memory limit (8 GB) and grab resource as fast as they can for efficiency.

Tested locally uploading two files each of 40 GB and maintain memory usage < 8 GB:
```
(.env) di@di-mbp ~/xetdata/pyxet/python/pyxet % xet cp ~/tt/zz* xet://localhost:3000:seanses-local/empty/main
copy /Users/di/tt/zz0, /Users/di/tt/zzx... to xet://localhost:3000:seanses-local/empty/main: (2 / 2), 80 GiB | 1.19 GiB/s, done.
```

### DIFF
--- a/rust/gitxetcore/src/constants.rs
+++ b/rust/gitxetcore/src/constants.rs
@@ -56,3 +56,7 @@ pub const XET_BACKUP_COMMIT_EMAIL: &str = "system@xethub.com";
 
 // Approximately 4 MB min spacing between global dedup queries.  Calculated by 4MB / TARGET_CHUNK_SIZE
 pub const MIN_SPACING_BETWEEN_GLOBAL_DEDUP_QUERIES: usize = 256;
+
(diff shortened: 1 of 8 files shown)


hoytak:
So...  This seems really complicated, and I'm worried there will be side effects if it's not used correctly...  Since we're using fixed sized buffers, I think we could get around this by simply enforcing a maximum buffer size globally, then calculating the size of the queue in order to respect the larger size limits wrt max buffer size.

seanses:
After offline discussion, close in favor of https://github.com/xetdata/xet-core/pull/399 for unforeseen risk.

PR #371, opened 2024-05-16
Fix for build break involving new git version. 

hoytak:
As a security measure, hooks can be problematic.  This prevents this.


port8080:
Tests are still failing with similar errors, maybe that env var needs to be set in the GH runner config

Also, putting the following links here for context
https://lore.kernel.org/git/20240514181641.150112-1-sandals@crustytoothpaste.net/T/#t
https://github.blog/2024-05-14-securing-git-addressing-5-new-vulnerabilities/

ylow (approved):

PR #373, opened 2024-05-19
Show server message nicely on mount

seanses:
Fix Fun-89.

Test locally, note that only the server reply "repo.access_denied_free_trial%!(EXTRA string=http://localhost:3000///subscription/upgrade)" is printed.

- Mount
```
di@di-mbp ~/tt % git xet mount http://localhost:3000/seanses-local/pyxet.git
Mounting to "/Users/di/tt/pyxet"
repo.access_denied_free_trial%!(EXTRA string=http://localhost:3000///subscription/upgrade)
Error : failed to mount
```

- Writable Mount
```
di@di-mbp ~/tt % git xet mount http://localhost:3000/seanses-local/pyxet.git --writable
Writable mounts are currently an EXPERIMENTAL feature. Be warned. You might lose data!!
Known issues: Performance is poor.

Raw clone at "/Users/di/tt/pyxet_raw".
Mounting at "/Users/di/tt/pyxet".

You can access and make arbitrary modification in the mounted path and changes
will immediately reflect in the git state in the raw clone path.

Similarly you can perform git operations in the raw clone path and it will immediately
reflect in the mounted path. All git operations should work as expected.

If you use a git UI, point it to the raw path.

repo.access_denied_free_trial%!(EXTRA string=http://localhost:3000///subscription/upgrade)
Error : failed to mount
```

To compare, this is clone
```
di@di-mbp ~/tt % git xet clone http://localhost:3000/seanses-local/pyxet.git
Preparing to clone Xet repository.
Cloning into 'pyxet'...
remote: repo.access_denied_free_trial%!(EXTRA string=http://localhost:3000///subscription/upgrade)
fatal: unable to access 'http://localhost:3000/seanses-local/pyxet.git/': The requested URL returned error: 403
```

### DIFF
--- a/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
+++ b/rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
@@ -66,40 +66,41 @@ pub fn clone_xet_repo_or_display_remote_error_message(
     base_dir: Option<&PathBuf>,
     allow_stdin: bool,
 ) -> Result<(String, Option<String>)> {
-    let ret = clone_xet_repo(
+    let (clone_ret, repo, branch) = clone_xet_repo_impl(
         config,
         git_args,
         no_smudge,
         base_dir,
-        false,
+        false, // no pass through leads to Captured result below
         allow_stdin,
-        true,
-    );
-    if ret.is_ok() {
-        return ret;
+        false, // don't check result internally, will check below
+    )?;
+
+    let CloneRet::Captured((status_code, _stdout, stderr)) = clone_ret else {
+        unreachable!();
+    };
+
+    // clone finished correctly
+    if let Some(0) = status_code {
+        return Ok((repo, branch));
     }
 
-    // Extract stderr message from the Err string, capture it as "all",
-    // if "remote: .*" exists in the stderr message (which is how git print
+    // If "remote: .*" exists in the stderr message (which is how git print
     // out the reply from server), capture the reply as "remote".
     //
-    // Print "remote" if it exists, otherwise print "all".
-    let regex =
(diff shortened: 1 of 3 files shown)


Saba9 (approved):

rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
```diff
@@ -57,6 +57,47 @@ pub fn normalize_repo_path(path: &str) -> String {
...
+    if ret.is_ok() {
+        return ret;
+    }
+
+    // Extract stderr message from the Err string, capture it as "all",
+    // if "remote: .*" exists in the stderr message (which is how git print
+    // out the reply from server), capture the reply as "remote".
+    //
+    // Print "remote" if it exists, otherwise print "all".
+    let regex =
```
hoytak, line 87: the clone_xet_repo is pretty much only used in mount and in git xet clone, with the passthrough flag being the only difference between the two.  It's easy code to split up, so I'd rather just have separate methods and have this method run run_git_captured without the check flag so it returns the stderr on failure.    This is pretty brittle.

hoytak: See like 151 below -- and just duplicate that, but turn off the check flag so it only errors out for non-git related errors.

hoytak (changes requested):
Just the regex parsing something we already have is awkward....

rust/gitxetcore/src/git_integration/git_repo_plumbing.rs
seanses: Gotcha

seanses: Fixed to use the captured stderr.

hoytak (approved):
LGTM!

PR #375, opened 2024-05-21
Add ability to bypass staging during git xet repo migrate.

hoytak:
This PR adds the ability to bypass staging during git xet repo migrate, uploading new data to CAS directly. 

To use this, run with `XET_CAS_BYPASS_STAGING=1`

```
XET_CAS_BYPASS_STAGING=1 git xet repo migrate
```

### DIFF
--- a/rust/cas_client/src/staging_client.rs
+++ b/rust/cas_client/src/staging_client.rs
@@ -68,6 +68,13 @@ impl StagingClient {
     }
 }
 
+fn cas_staging_bypass_is_set() -> bool {
+    // Returns true if XET_CAS_BYPASS_STAGING is set to something besides "0"
+    std::env::var_os("XET_CAS_BYPASS_STAGING")
+        .filter(|v| v != "0")
+        .is_some()
+}
+
 /// Creates a new staging client wraping a staging directory.
 /// If a staging directory is provided, it will be used for staging.
 /// Otherwise all queries are passed through to the remote directly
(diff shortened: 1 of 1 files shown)


rust/gitxetcore/src/data/data_processing_v2.rs
```diff
@@ -807,14 +807,29 @@ impl PointerFileTranslatorV2 {
...
 
-            self.cas
-                .put(
-                    &self.prefix,
-                    &cas_hash,
-                    take(&mut cas_data.data),
-                    chunk_boundaries,
-                )
-                .await?;
+            if std::env::var_os("XET_CAS_BYPASS_STAGING").unwrap_or_default() == "1" {
```
assafvayner, line 810: minute nit, I'd just check for existing rather than the value == 1, you should be able to just check for `.is_some()`

assafvayner: or `is_ok()`

seanses (commented):
Instead of low level changes, probably better and easier to do it from config like in https://github.com/xetdata/xet-core/blob/dd4af199c5b1f932174fcf19ab3e2088e3e4db77/rust/gitxetcore/src/xetblob/xet_repo.rs#L205

rust/gitxetcore/src/data/data_processing_v2.rs
hoytak: What we've been doing up to now is have "0" explicitly turn it off.  It's a convention, but it allows us to override things in subprocesses if need be.  Not that that would happen here, but I like consistency.

hoytak:
> Instead of low level changes, probably better and easier to do it from config like in
> 
> https://github.com/xetdata/xet-core/blob/dd4af199c5b1f932174fcf19ab3e2088e3e4db77/rust/gitxetcore/src/xetblob/xet_repo.rs#L205

Doesn't this just cause it to use a tmp directory instead of bypass staging altogether?  Let me look...

seanses:
> Doesn't this just cause it to use a tmp directory instead of bypass staging altogether? Let me look...

No.. this makes it use PassthroughStagingClient: https://github.com/xetdata/xet-core/blob/dd4af199c5b1f932174fcf19ab3e2088e3e4db77/rust/cas_client/src/staging_client.rs#L97

PR #378, opened 2024-05-29
Cache CAS endpoint query

hoytak:
This PR caches the CAS endpoint query across runs.  With this PR, the endpoint should be queried at most once every 5 min or so.  Helps with git xet repo migrate.


rust/gitxetcore/src/environment/query_cache.rs
```diff
@@ -0,0 +1,215 @@
...
+        if let Ok(json_string) = serde_json::to_string_pretty(&qv).map_err(|e| {
+            debug!("Error serializing query values to JSON: {:?}", e);
+            e
+        }) {
+            debug!(
+                "CachedQueryInfo:save: Serializing cached query struct {qv:?} to {:?}",
+                &self.file_name
+            );
+
+            let _ = std::fs::write(&self.file_name, json_string).map_err(|e| {
```
seanses, line 141: This is under `xet_home`, so ideally `Permission::create_file` before `std::fs::write`. As a side issue, in the future we may want to make `xet_home` a special object than PathBuf so it handles write with permission internally.

seanses (approved):
LGTM! 

Additional request: since there's a plan to run `repo migrate` on Xetea servers, can we pass down `--cas` to `git xet smudge` here: https://github.com/xetdata/xet-core/blob/992fbf203a4c2cb9bd8dbe81f1f732537e2a62eb/rust/gitxetcore/src/git_integration/repo_migration.rs#L737

rust/gitxetcore/src/environment/query_cache.rs
hoytak: It already creates the query_cache directory using Permission, so I'm not as worried about it here.

PR #379, opened 2024-05-29
Improve merkledb load speed for non-dedup tasks

hoytak:
This PR makes the smudge path far more efficient, especially when there are many merkledb shards.  
- Chunk lookup table is loaded lazily on chunk dedup query; when chunk dedup query doesn't happen, it's never loaded.
- Shard IO is done async in parallel now.


hoytak:
Have another plan for this, holding back....

PR #381, opened 2024-05-29
Convenience script for updating the version

hoytak:
There have been several issues with the version changing code currently in use: 
- version of a dependency gets updated if specified using some forms of Cargo.toml syntax.
- libxet/Cargo.lock keeps not getting updated. 

In addition, convenience scripts are convenient. 

This PR introduces a script, run as ./scripts/update_version.sh, that does all the version updating correctly.

### DIFF
--- a/scripts/update_version.sh
+++ b/scripts/update_version.sh
@@ -1,0 +1,84 @@
+#!/bin/bash -e 
+
+# Function to check if the version string is valid
+is_valid_version() {
+    local version=$1
+    if [[ $version =~ ^[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+print_help() {
+    echo "Usage: $0 <new version>"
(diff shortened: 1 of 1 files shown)


scripts/update_version.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# ensure main is clean
+git status --porcelain
+
+git checkout -b update-version-$XET_NEXT_VERSION
+
+# Update project Cargo.toml files and Cargo.lock file
+# Replace only the `version = "..."` string in the first 5 lines to avoid 
+# replacing the version line for a dependency.
+find . -name 'Cargo.toml' | xargs sed -i '' "1,5s/^version = \"${current_version}\"$/version = \"$XET_NEXT_VERSION\"/"
```
seanses, line 71: This will update https://github.com/xetdata/xet-core/blob/main/rust/xet_error/impl/Cargo.toml, minor concern

seanses (approved):

PR #382, opened 2024-05-30
"merkledb extract-git" fetch no shard by default

seanses:
Xetea [GetXetDirSummary](https://github.com/xetdata/xethub/blob/13542f1cdb7a7d6c4de38e1b1b0827c51dc70194/golang/xetea/modules/context/xet.go#L83) and [GetDiff](https://github.com/xetdata/xethub/blob/13542f1cdb7a7d6c4de38e1b1b0827c51dc70194/golang/xetea/services/gitdiff/gitdiff.go#L1274) call `merkledb extract-git` for MDBV1 compatibility. This PR disables fetching all shards by default for V2 repos so there's less burden on Xetea. Fix [CLI-224](https://linear.app/xethub/issue/CLI-224/merkledb-extract-git-takes-long-time-when-rendering-pr-for-the-real)

### DIFF
--- a/rust/gitxetcore/src/command/merkledb.rs
+++ b/rust/gitxetcore/src/command/merkledb.rs
@@ -155,6 +155,9 @@ struct MerkleDBGitExtractArgs {
     output: Option<PathBuf>,
     #[clap(short, long, default_value = "refs/notes/xet/merkledb")]
     notesref: String,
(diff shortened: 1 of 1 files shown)


hoytak (approved):
LGTM for this limited context.

PR #383, opened 2024-06-06
Remove git notes caching for dir summaries

znation:
Now that dir summaries are cached server side, caching in git notes is redundant. Removing it should also help mitigate some random lockfile related bugs we've seen related to the dir-summary notes lockfiles.


seanses (approved):
LGTM! "cached server side" meaning in Xet-Meta or in xetea cache (either in-memory or redis)?

znation:
Server side in this case meaning Xetea cache (redis).

PR #386, opened 2024-06-09
Adds download progress reporting for xet. 

hoytak:
This PR adds the xet-core half of the download progress reporting to the direct read-to-file path.  It also renames get(path) to read_to_path(path), which is much less confusing.  

Note: this is tested, but the testing is done with a local version of pyxet, which is up at https://github.com/xetdata/pyxet/pull/140.


seanses (approved):
LGTM!

PR #387, opened 2024-06-11
Lower the number of default concurrent upload/download

seanses:
Fix CLI-221

- lower default concurrent upload/download to 8
- provide env var to reset the default


hoytak (approved):
Lgtm!

PR #395, opened 2024-06-19
Fix for windows integration test issues.

hoytak:
So it turns out that canonicalize() is broken on windows in the sense that a canonicalized path can no longer be treated as a linux based path due to //?/ being prepended to it.  This fixes this issue.


rust/gitxetcore/Cargo.toml
```diff
@@ -100,6 +100,7 @@ blake3 = "1.5.1"
 uuid = { version = "1.8.0", features = ["std", "rng", "v6"] }
 lz4 = "1.24.0"
 git-url-parse = "0.4.4"
+path-absolutize = "3.1.1" # Can drop after rust 1.79
```
seanses, line 103: What will change after 1.79?

seanses (approved):
LGTM!

PR #396, opened 2024-06-19
More fixes for note conversion path in repo migration. 

hoytak:
Currently, the note conversion ran as a separate stage after the regular conversion, in which all blob oids in that stage reachable by notes and their dependents were considered note blobs.   The exception to the building of this dependency tree is that attached commits that were already converted would be ignored.   

The problem is that many blobs in this path may not actually be note blobs and should be put through the filter.  The cases missed: 
- notes attached to blobs. 
- dependents of notes attached to trees.  (notes have to be attached to a previously converted commit to turn off the switch saying it's a note). 

This approach tracks whether an oid can be considered as a note through the different stages so blobs are filtered and not filtered properly.


rust/gitxetcore/src/git_integration/repo_migration/migration.rs
```diff
@@ -780,17 +775,14 @@ pub async fn migrate_repo(
                 match obj.kind().unwrap_or(ObjectType::Any) {
                     ObjectType::Tree => {
                         let deps = {
-                            if in_note_conversion_stage
-                                && commit_tree_oids.contains(&oid)
-                                && !known_non_note_oids.contains(&oid)
-                            {
+                            if in_note_conversion_stage && is_note_oid {
                                 get_note_tree_dependents(&src, obj, &full_tr_map)
```
hoytak, line 779: Actually, tree OIDs in the tree field of a note tree is itself a note tree.   Just the OIDs in the path field are not.

seanses (approved):

PR #404, opened 2024-06-28
File utilities needed for XetFDLS

hoytak:
This PR adds a number of file utilities that robustly handles file creation with the correct permissions and metadata copying so files match.  This is used by the XetLDFS utility, but pulls these functions out of the config code and into standalone utilities.


rust/Cargo.toml
```diff
@@ -25,8 +25,13 @@ members = [
     "lazy",
     "error_printer",
     "tableau_summary",
-    "chunkpipe"]
+    "chunkpipe",
+    "file_utils",
+]
 
+[pofile.debug]
```
seanses, line 32: profile

seanses (approved):

PR #410, opened 2024-07-12
XetLDFS: Improved error reporting macros.

hoytak:
First step in getting better error reporting: 

ld_error! -- functions just like error!, but dumps to stderr if the error subscribing hasn't been initialized. 
ld_warn! -- same thing.

Also moved the macros into their own module for cleanliness and made sure each was self-contained.


seanses (approved):
LGTM!

PR #412, opened 2024-07-16
Cleanup path logic; initialize the xet repos from path variable.

hoytak:
This PR simplifies and cleans up the path logic in the LDFS repo initialization and detection of which repositories get loaded and when.  After this PR, we:
- Use impl AsRef<Path> consistently for main functions and then have unsafe _c versions that take paths as *const c_char. 
- Repository locations are loaded from environment variables.
- Error messages are logged and handled consistently.


xetldfs/src/xet_interface.rs
```diff
@@ -1,194 +1,163 @@
...
-        *cfg_wrap = Some(cfg);
-    }
+    for mut path in paths {
+        // Trim out the whitespace.
+        path = path.trim();
 
-    Ok(cfg_wrap.as_ref().unwrap().clone())
-}
+        // Is the path quoted?
+        if path.len() >= 2
```
seanses, line 47: Just realized that *unix allow all special characters in a path component except for `/`, this means that `"a.txt"` is actually a valid path name. And since shells remove quotes typed in the command line unless users deliberately want them in by escaping the quotes, we should treat them as users' intention.

hoytak: Ah, gotcha, good point.  I guess the user would quote the entire environment variable string in this case.

xetldfs/tests/integration_tests/test_python.sh
```diff
@@ -23,7 +23,7 @@ done
 
 # pandas Read
 (
-    xetfs_on
+    xetfs_on ./
```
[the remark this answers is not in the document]
hoytak: Good catch

seanses (approved):
LGTM!

PR #413, opened 2024-07-19
Updated stat interposing

hoytak:
This PR:
- Imports the hook and real macros, modifying them on linux to crash not at initialize time but at runtime when a function that doesn't exist is possibly interposed. 
- Adds full coverage for all the libc and glibc stat functions and variants, also consolidating the stat implementation to a single easier-to-read function.
- Use statx on linux to determine regular files.


seanses (approved):
LGTM!

xetldfs/src/stat_interposing.rs
```diff
@@ -0,0 +1,342 @@
...
+                    ld_error!("Unknown function stat");
+                    errno::set_errno(errno::Errno(ENOSYS)); // Function does not exist
+                    return -1;
+                }
+            }
+        })
+    }
+}
+
+// int __xstat(int ver, const char *pathname, struct stat *buf);
```
seanses, line 115: Document for all the `__x*` that ver should be 3.

PR #416, opened 2024-07-22
Replace snailquote with shellish due to licensing

port8080:

PR #417, opened 2024-07-23
Fix for file permissions testing issue.

hoytak:
The current test for a file permission issue assumed a default which is not actually true on all platforms.  This removes that check. 

Also, an unnecessary reloading of file permissions happens on the close, which gets removed as well.


rust/gitxetcore/Cargo.toml
```diff
@@ -79,7 +79,7 @@ nfsserve = "0.10"
 atty = "0.2"
 libc = "0.2"
 itertools = "0.10.5"
-shellish_parse = "2.2"
+snailquote = "0.3.1"
```
seanses, line 82: Likely an incorrect merge happened, this reverted https://github.com/xetdata/xet-core/pull/416

PR #419, opened 2024-08-29
refactor data processing

seanses:


ylow:
yikes. This will conflict with lots of stuff :-)

